PR #18, opened 2026-04-03
fix: dashboard activity chart flatlines on sessions

dontcallmejames:
## Problem

The activity chart on the dashboard shows no useful data — sessions flatlines at zero and messages appears empty even with an active Hermes gateway.

Three root causes:

### 1. `toSessionSummary()` drops message/tool counts

`message_count` and `tool_call_count` exist on the gateway's `/api/sessions` response but were never mapped into the session summary object returned by the workspace proxy. Every session reaches the chart with `0` messages.

### 2. Single Y-axis with incompatible scales

Messages count in the hundreds; sessions count in single digits. On a shared Y-axis, sessions renders as a flat line at the bottom of the chart.

### 3. Cache collision with chat sidebar

The dashboard query used `chatQueryKeys.sessions` — the same React Query key as the chat sidebar. The sidebar fetches fewer sessions and overwrites the dashboard's larger dataset, causing stale or missing data.

## Fix

- Add `message_count` and `tool_call_count` to `toSessionSummary()`
- Dual Y-axes: left axis (green) for messages, right axis (purple) for sessions
- Dedicated `['dashboard', 'sessions']` query key, separate from the chat sidebar
- Trim leading empty days from the 14-day window so the chart fills with actual activity

### DIFF
--- a/src/screens/dashboard/dashboard-screen.tsx
+++ b/src/screens/dashboard/dashboard-screen.tsx
@@ -15,7 +15,6 @@ import type { HermesSession } from '@/server/hermes-api'
 import { chatQueryKeys } from '@/screens/chat/chat-queries'
 import { getCapabilities } from '@/server/gateway-capabilities'
 import { useIsFeatureAvailable } from '@/hooks/use-gateway-caps'
-import type { HermesSession } from '@/server/hermes-api'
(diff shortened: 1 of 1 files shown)


dontcallmejames:
@outsourc-e will do. I'll work on it tonight.

PR #22, opened 2026-04-04
fix: dashboard activity chart dual Y-axis and leading empty day trim

dontcallmejames:
Follow-up to #18 (now closed) — contains only the chart improvements as requested.

## Changes
- Dual Y-axis on the activity chart (sessions on left, tool calls on right) so both metrics are readable at their own scale
- Trims leading empty days from the chart window so the graph doesn't start with a flat line when the instance is new

## Notes
- Rebased on latest main
- Updated capability gating to use `useFeatureAvailable` from `@/hooks/use-feature-available` (aligning with #19)
- Removed old `useIsFeatureAvailable` import from `use-gateway-caps`

### DIFF
--- a/src/screens/dashboard/dashboard-screen.tsx
+++ b/src/screens/dashboard/dashboard-screen.tsx
@@ -11,10 +11,9 @@ import {
   ResponsiveContainer,
   CartesianGrid,
 } from 'recharts'
-import { listSessions, getConfig } from '@/server/hermes-api'
+import type { HermesSession } from '@/server/hermes-api'
 import { chatQueryKeys } from '@/screens/chat/chat-queries'
 import { getCapabilities } from '@/server/gateway-capabilities'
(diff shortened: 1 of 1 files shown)


dontcallmejames:
Refreshed this branch against current upstream/main and trimmed overlap that upstream already merged.\n\nThe remaining diff is intentionally narrow: dashboard chart behavior only (dual Y-axis, client-side proxy fetch path, dedicated query key / trimmed leading empty days), with no extra hermes-api overlap left in the branch.

dontcallmejames:
Refreshed this branch against current upstream/main and trimmed overlap that upstream had already merged.

The remaining diff is intentionally narrow:
- dashboard chart behavior only
- dual Y-axis
- client-side proxy fetch path
- dedicated query key
- trimmed leading empty days

No extra hermes-api overlap remains in the branch.

PR #53, opened 2026-04-13
fix(tasks): constrain kanban board to page width + eliminate refetch flicker

dontcallmejames:
## Changes

**Width fix:**
- Add `max-w-[1200px] mx-auto` to board container to match header constraint
- Reduce column width from 280px to 210px so 5 columns fit under 1200px without horizontal overflow
- Reduce column min-width from 240px to 200px

**Refetch flicker fix:**
- Add `placeholderData: keepPreviousData` to keep stale cards visible during 30s background refetch
- Show skeletons only on true initial load (`data === undefined`), not on background refetch
- Fix column count to show task count during refetch instead of '…' flicker

## Before vs After

**Width:** 5×280px columns + gaps overflowed ~1280px viewports → now 5×210px columns fit comfortably under 1200px

**Flicker:** Every 30s the board would blank out with skeleton cards during background refetch → now stale data stays visible until fresh data arrives

Closes: kanban columns overflow page causing nested scrollbars  
Closes: 30s background refetch causes skeleton flicker

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,6 +34,8 @@
     "@types/react-grid-layout": "^1.3.6",
     "class-variance-authority": "^0.7.1",
     "clsx": "^2.1.1",
+    "cytoscape": "^3.33.2",
+    "cytoscape-fcose": "^2.2.0",
     "marked": "^17.0.1",
     "motion": "^12.29.2",
     "playwright": "^1.58.2",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -56,6 +56,12 @@ importers:
       clsx:
         specifier: ^2.1.1
         version: 2.1.1
+      cytoscape:
+        specifier: ^3.33.2
+        version: 3.33.2
+      cytoscape-fcose:
+        specifier: ^2.2.0
+        version: 2.2.0(cytoscape@3.33.2)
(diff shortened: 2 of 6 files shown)


outsourc-e (commented):
Thanks for this — the kanban width fix and refetch flicker fix are solid improvements.

However, this PR bundles two unrelated changes:

1. **Kanban width + refetch flicker fix** — the actual fix described in the title
2. **WikiGraph component** — a new feature with 2 new npm dependencies (cytoscape, cytoscape-fcose) and ~400 lines of new code

Could you split this into two separate PRs? The kanban fix is ready to merge on its own. The WikiGraph feature deserves its own review since it adds new dependencies and a significant new component.

Also a minor note on the WikiGraph code: `require('cytoscape-fcose')` is a CJS require — would be cleaner as a dynamic `import()` or ESM import to match the rest of the codebase.

Happy to merge the kanban fix quickly once it's isolated. 🙏

dontcallmejames:
I jumped the gun on this PR, I apologize. I'll definitely split them into two PR's. I'm also going to make a small adjustment to the Kanban width fix. I'm not sure why that wikigraph component got bundled with this. It was something else I was working on, and should not have been included.

PR #54, opened 2026-04-13
fix(tasks): dynamic column widths + direct file API (no gateway dependency)

dontcallmejames:
### Kanban UI fixes

**Column width fix (tasks-screen.tsx):**
- Calculate `colMaxWidth = Math.floor(1200 / visibleColumns.length)` dynamically
- 4 columns (done hidden) → 300px each, fills the 1200px container
- 5 columns (done shown) → 240px each, fits without overflow

**Error state fix (tasks-screen.tsx):**
- Add `isError` check — shows 'Failed to load tasks' with Retry button instead of cycling through skeleton cards indefinitely
- Keeps `keepPreviousData` for refetch, adds explicit `isLoading` for initial load

### Tasks API fix

**Direct file API (hermes-tasks.ts):**
- Replaced proxy-based API (`HERMES_API/api/tasks`) with direct `~/.hermes/tasks.json` reads/writes
- Tasks API now works without requiring the gateway to be running
- GET, POST, PATCH, DELETE all handled locally

**Proxy removal (vite.config.ts):**
- Removed the `/api/hermes-tasks` → `/api/tasks` proxy rule (no longer needed)

## Scope
- `src/screens/tasks/tasks-screen.tsx` — UI fixes (1 file)
- `src/routes/api/hermes-tasks.ts` — file-based API (new implementation)
- `vite.config.ts` — proxy cleanup
- No dependency changes, no new packages

Closes: kanban columns overflow page (4 vs 5 column layout)
Closes: infinite skeleton cycle when tasks API errors
Closes: tasks API broken without gateway running

### DIFF
--- a/src/routes/api/hermes-tasks.ts
+++ b/src/routes/api/hermes-tasks.ts
@@ -1,31 +1,220 @@
+/**
+ * Tasks API — JSON file persistence layer.
+ * Reads/writes ~/.hermes/tasks.json (simple, portable, no native modules needed in SSR).
+ */
 import { createFileRoute } from '@tanstack/react-router'
 import { isAuthenticated } from '../../server/auth-middleware'
-import { HERMES_API } from '../../server/gateway-capabilities'
+import { existsSync, readFileSync, writeFileSync, mkdirSync } from 'node:fs'
+import { resolve } from 'node:path'
+import os from 'node:os'
+
+const HERMES_HOME = process.env.HERMES_HOME ?? resolve(os.homedir(), '.hermes')
+const TASKS_FILE = resolve(HERMES_HOME, 'tasks.json')
+
+interface Task {
+  id: string
+  title: string
+  description: string
+  column: string
+  priority: string
+  assignee: string | null
+  tags: string[]
+  due_date: string | null
+  position: number
+  created_by: string
+  created_at: string
+  updated_at: string
+}
+
+function ensureFile() {
+  const dir = HERMES_HOME
+  if (!existsSync(dir)) {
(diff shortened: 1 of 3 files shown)


outsourc-e (commented):
Reviewed. There are good pieces here that we should keep, but I would not merge this backend change as-is.\n\nWhat looks good:\n- dynamic Kanban column widths\n- explicit error state + retry for tasks\n- keepPreviousData anti-flicker improvement\n\nWhat needs adjustment before merge:\n- the new local tasks API changes the mutation response contract: the client currently expects  for create/update/move, but this route returns the raw task object\n- the route split is incomplete:  becomes local, but  is still gateway-backed, so “no gateway dependency” is not actually true yet\n- delete/update/move semantics no longer line up cleanly with the current frontend client usage\n\nSuggested path:\n- keep the UI improvements\n- either keep tasks gateway-backed for now, or finish the local file-backed implementation across all task routes while preserving the current frontend contract\n\nSo: good contribution, worth salvaging, but not mergeable raw without a small follow-up cleanup.

outsourc-e:
Follow-up clarification on my review since GitHub ate the inline code formatting: the client currently expects `{ task: ... }` responses for create/update/move. In this PR, `/api/hermes-tasks` becomes local/file-backed, but `/api/hermes-tasks/$taskId` is still gateway-backed. That split is the main reason I said not to merge the backend portion raw.

outsourc-e:
Merged a cleaned-up follow-up in #57 that keeps the UI wins from this PR and completes the local task API without the contract drift. Closing this one to avoid duplicate paths — thanks for the contribution.

PR #79, opened 2026-04-20
fix: harden workspace installer for local Hermes API

Hamoun-IA:
## Summary
- make `HERMES_API_URL` updates idempotent instead of append-only
- enable `API_SERVER_ENABLED=true` in the Hermes env so the local workspace can actually reach port 8642
- fail clearly if `INSTALL_DIR` exists but is not a git repo
- remove the bogus `pnpm start:all` guidance and clarify the restart requirement

## Why
The current installer leaves a fresh local setup in a half-working state unless the user already knows the Hermes HTTP API is opt-in. It also advertises a non-existent `pnpm start:all` command.

## Validation
- `bash -n install.sh`
- inspected `package.json` to confirm `pnpm start:all` does not exist
- checked `.env.example` comment about `API_SERVER_ENABLED=true` and aligned installer behavior with it

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -7,9 +7,9 @@
 # What it does:
 #   1. Verifies Node 22+, git, pnpm
 #   2. Installs hermes-agent via Nous's official upstream installer
-#      (hermes-agent is NOT on PyPI; it's a source install handled by Nous)
 #   3. Clones hermes-workspace
-#   4. Sets up .env, installs deps, links bundled skills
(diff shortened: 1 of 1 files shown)


outsourc-e:
Merged — thanks for the clean diff. All four fixes were things I'd been meaning to harden (especially the `API_SERVER_ENABLED=true` step; a lot of fresh installs were hitting that silently). Appreciate the contribution 🙌

PR #81, opened 2026-04-21
fix(auth): persist session tokens to ~/.hermes/workspace-sessions.json

Klng79:
Replace in-memory Set with a file-backed store so sessions survive server restarts. Tokens have a 30-day TTL with automatic expiration pruning. File location honours HERMES_HOME env var.

### DIFF
--- a/src/server/auth-middleware.ts
+++ b/src/server/auth-middleware.ts
@@ -1,10 +1,89 @@
 import { randomBytes, timingSafeEqual } from 'node:crypto'
+import { existsSync, readFileSync, writeFileSync } from 'node:fs'
+import { join } from 'node:path'
 
 /**
- * In-memory session store.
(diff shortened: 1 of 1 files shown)


outsourc-e (commented):
Solid fix — clean scoping, graceful corruption handling, honors `HERMES_HOME`, and the load-time pruning is a nice touch.

One small nit before merge: the inline `const { mkdirSync } = require('node:fs')` in `saveStore()` is an ESM/CJS mixup — this file already uses `import` at the top. Could you hoist it into the existing import?

```ts
import { existsSync, mkdirSync, readFileSync, writeFileSync } from 'node:fs'
```

Happy to fix-forward myself if easier. Nothing else blocking from my side. 🙏

Klng79:
yes please fix-forward

PR #93, opened 2026-04-21
docs: clarify zero-fork attach requirements

agent-cortex:
## Summary
- document that zero-fork workspace attachments need both the Hermes gateway (`:8642`) and the Hermes dashboard (`:9119`)
- add `HERMES_DASHBOARD_URL` to the existing-agent attach example
- clarify that `API_SERVER_ENABLED=true` only covers the gateway/core APIs
- add explicit verification steps for both `/health` and `/api/status`
- fix the quickstart section so it starts `hermes dashboard` alongside `hermes gateway run`

## Why
We hit a real docs/runtime mismatch during setup. The README implied that attaching the workspace to an already-running Hermes install only required `HERMES_API_URL` plus the gateway API server. In practice, this build runs in zero-fork mode and needs the separate dashboard API for configuration, sessions, skills, and jobs. Without that, the workspace falls back to portable mode and shows the misleading configuration error about missing extended APIs.

## Requirements / operator notes
- `hermes gateway run` serving core APIs on `:8642`
- `hermes dashboard` serving dashboard APIs on `:9119`
- optional `HERMES_API_TOKEN` only when the gateway uses `API_SERVER_KEY`
- `HERMES_DASHBOARD_URL` should be set explicitly when attaching to an existing install

## Verification
- `curl http://127.0.0.1:8642/health`
- `curl http://127.0.0.1:9119/api/status`
- start workspace with both `HERMES_API_URL` and `HERMES_DASHBOARD_URL` set and confirm zero-fork mode

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,11 +81,13 @@ cd hermes-workspace
 pnpm install
 cp .env.example .env
 
-# Point at your existing gateway. Local, Tailscale, LAN — whatever URL works.
+# Point at your existing Hermes services.
 echo 'HERMES_API_URL=http://127.0.0.1:8642' >> .env
+# Zero-fork installs also need the separate dashboard API for config/sessions/skills/jobs.
+echo 'HERMES_DASHBOARD_URL=http://127.0.0.1:9119' >> .env
 
 # If your gateway was started with API_SERVER_KEY (auth enabled), set the same value:
-# echo 'HERMES_API_TOKEN=<your-key>' >> .env
+# echo 'HERMES_API_TOKEN=***' >> .env
 
 pnpm dev                            # http://localhost:3000
 ```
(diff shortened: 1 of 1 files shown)


agent-cortex:
Real-world failure mode from setup on a Pi/Tailscale/systemd install:

We followed the existing "attach to already-running Hermes Agent" README path with `HERMES_API_URL=http://127.0.0.1:8642` and `API_SERVER_ENABLED=true`, then exposed the workspace on port 3002 behind Tailscale. Chat worked, but the workspace stayed in `portable` mode and Configuration failed with:

> Configuration requires a Hermes gateway that exposes the extended APIs. Check that Hermes is installed and running with `hermes gateway run`.

Root cause: the installed workspace build was actually running in zero-fork mode, so config/sessions/skills/jobs depended on the separate dashboard API on `:9119`, not just the gateway on `:8642`. Once `hermes dashboard` was started and `HERMES_DASHBOARD_URL=http://127.0.0.1:9119` was added, the workspace reprobed into `mode=zero-fork` and everything unlocked correctly.

So the docs change here is not theoretical — it fixes a concrete setup trap for existing Hermes installs, especially when users run the services under systemd and only verify the gateway health endpoint.

agent-cortex:
Real-world failure mode from setup on a Pi/Tailscale/systemd install:

We followed the existing "attach to already-running Hermes Agent" README path with `HERMES_API_URL=http://127.0.0.1:8642` and `API_SERVER_ENABLED=true`, then exposed the workspace on port 3002 behind Tailscale. Chat worked, but the workspace stayed in `portable` mode and Configuration failed with:

> Configuration requires a Hermes gateway that exposes the extended APIs. Check that Hermes is installed and running with `hermes gateway run`.

Root cause: the installed workspace build was actually running in zero-fork mode, so config/sessions/skills/jobs depended on the separate dashboard API on `:9119`, not just the gateway on `:8642`. Once `hermes dashboard` was started and `HERMES_DASHBOARD_URL=http://127.0.0.1:9119` was added, the workspace reprobed into `mode=zero-fork` and everything unlocked correctly.

So the docs change here is not theoretical — it fixes a concrete setup trap for existing Hermes installs, especially when users run the services under systemd and only verify the gateway health endpoint.

PR #132, opened 2026-04-23
feat(admin): gate Hermes Agent settings/providers behind SSO allowlist

svettypoo:
## Summary

- Adds `HERMES_ADMIN_EMAILS` env-driven admin gating so the workspace can sit behind S&T Properties SSO while still exposing chat to all staff and keeping Hermes Agent / provider configuration admin-only.
- New server helper (`src/server/admin-auth.ts`) reads forwarded SSO identity headers (`X-Forwarded-User`, `X-SSO-Email`, `X-Forwarded-Email`, plus a few common variants) and matches the email against a comma-separated allowlist. Case-insensitive. Empty allowlist preserves the previous "everyone is admin" behavior for local / single-user setups.
- New `/api/me` endpoint returns `{ email, isAdmin, adminAllowlistConfigured }` and is consumed by a `useCurrentUser` React-Query hook.
- Route guards on `/settings/providers` and `/settings/mcp` render a friendly `AdminOnlyState` (with a link back to chat) instead of the generic "backend unavailable" prompt.
- `SettingsSidebar`, `SettingsMobilePills`, and the `SettingsDialog` (mini quick-settings) hide admin-only sections — Connection, Model & Provider, Agent Behavior, Smart Routing, Voice, Display, MCP. Non-admins still see Appearance, Chat, Notifications, Language. Direct-URL access to an admin section (e.g. `?section=hermes`) renders `AdminOnlyState` in-place.
- Model / provider catalogue entries are untouched; only access to the settings surface changes.

## Deployment

Set the allowlist on the Hetzner deployment's env:

```
HERMES_ADMIN_EMAILS=pargov.svet@gmail.com,Tina@stproperties.org
```

Caddy already forwards SSO identity as `X-Forwarded-User` (derived from `X-SSO-Email`) — no infra change required.

## Test plan

- [x] `npm test` — 31 tests pass (including 4 new tests covering allowlist parsing and identity header extraction)
- [x] `npm run build` — production build succeeds
- [x] `npx tsc --noEmit` — no new type errors (one pre-existing error in `src/components/ui/input.tsx` is unrelated)
- [ ] Deploy and verify: non-admin sees chat + non-admin settings sections, admin-only sections and `/settings/providers` render "Admin only"; admins see the full settings surface as before.

### DIFF
--- a/src/components/admin-only-state.tsx
+++ b/src/components/admin-only-state.tsx
@@ -1,0 +1,50 @@
+import { Link } from '@tanstack/react-router'
+import { HugeiconsIcon } from '@hugeicons/react'
+import { LockIcon } from '@hugeicons/core-free-icons'
+import { Button } from '@/components/ui/button'
+import { useCurrentUser } from '@/hooks/use-current-user'
+
+type Props = {
+  featureLabel?: string
+}
+
+export function AdminOnlyState({
+  featureLabel = 'This area',
+}: Props) {
+  const { email } = useCurrentUser()
+
+  return (
+    <div className="flex min-h-[60vh] w-full items-center justify-center p-6">
+      <div className="flex w-full max-w-md flex-col items-center gap-4 rounded-2xl border border-primary-200 bg-primary-50/80 p-8 text-center shadow-sm backdrop-blur-xl">
+        <span className="inline-flex size-12 items-center justify-center rounded-2xl border border-primary-200 bg-primary-100/70">
+          <HugeiconsIcon icon={LockIcon} size={24} strokeWidth={1.5} />
+        </span>
+        <div className="space-y-1">
+          <h2 className="text-lg font-semibold text-primary-900">
+            Admin only
+          </h2>
+          <p className="text-sm text-primary-600 text-pretty">
+            {featureLabel} is restricted to Hermes administrators. Ask an admin
+            to adjust the <code className="font-mono text-xs">HERMES_ADMIN_EMAILS</code>{' '}
+            allowlist if you need access.
+          </p>
+          {email ? (
+            <p className="text-xs text-primary-500">
+              Signed in as <span className="font-mono">{email}</span>
+            </p>
+          ) : null}
+        </div>
+        <div className="flex gap-2">
+          <Link to="/chat/$sessionKey" params={{ sessionKey: 'main' }}>
+            <Button size="sm">Back to chat</Button>
+          </Link>
+          <Link to="/dashboard">
+            <Button size="sm" variant="outline">
+              Dashboard
+            </Button>
+          </Link>
+        </div>
+      </div>
+    </div>
+  )
+}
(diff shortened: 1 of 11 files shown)


outsourc-e:
Thanks @svettypoo — the SSO gating concept is clean and the test coverage on `admin-auth.ts` is solid. Three items before we merge:

**1. Server-side enforcement missing (blocker).** Right now the gating is UI-only — route components render `<AdminOnlyState/>` for non-admins, but the underlying API endpoints (e.g. `/api/config`, `/api/providers`, `/api/mcp`, `/api/hermes`) will happily accept mutations from a non-admin who crafts a direct request (browser console, curl, any API client). Please wrap each mutation route with `isAdminRequest(request)` and return 403 when false. Otherwise the feature is security theater.

**2. Header trust (needs explicit note).** The `IDENTITY_HEADERS` list accepts `x-forwarded-user`, `x-sso-email`, etc. unconditionally. When the workspace isn't behind a sanitizing reverse proxy, any client can send those headers and claim any identity. We just merged #133 which adds `TRUST_PROXY` for related reasons — please either:
   - make admin-header trust conditional on `TRUST_PROXY=1` being set (safer default), or
   - document clearly in the PR body + a README note that `HERMES_ADMIN_EMAILS` must only be set when a sanitizing proxy is in front of the workspace.

**3. Fail-open default (tradeoff).** `isAdminEmail(email)` returns `true` when `HERMES_ADMIN_EMAILS` is unset. That's the right call for single-user local installs, but please add a startup log line (`[admin-auth] HERMES_ADMIN_EMAILS not set — all users are admin`) so nobody accidentally deploys in fail-open mode to a shared SSO tenant. Better yet, consider fail-closed if both `HERMES_PASSWORD` and `HERMES_ADMIN_EMAILS` are unset and `HOST` is non-loopback — that combination on a shared host is the actual danger state.

Happy to merge once (1) lands and (2) is clarified. (3) is a nice-to-have not a blocker.

outsourc-e (changes requested):
Thanks for pushing this — the UX direction is useful, but I don't think we can merge it yet because it only gates the settings UI.

Must fix before merge:

1. Add server-side authorization to the admin/config mutation/read endpoints, not just `/api/me` + route hiding. A non-admin can still call the underlying settings/provider/MCP/config APIs directly if they know the route. The allowlist check needs to live on every route that exposes or mutates Hermes Agent/provider/backend config.

2. Be explicit about trust boundaries for forwarded identity headers. `X-Forwarded-User`, `X-SSO-Email`, etc. are only trustworthy if this process is guaranteed to sit behind the SSO proxy and direct access is impossible. Otherwise a client can spoof those headers. Please either gate header trust behind deployment config / trusted proxy assumptions, or document/enforce the expected proxy-only topology.

3. Reconsider fail-open semantics. The PR says empty `HERMES_ADMIN_EMAILS` preserves “everyone is admin”; that may be fine for local dev, but it is dangerous in production. At minimum this should be an explicit dev/default mode with clear deployment docs, not an accidental unset-env = admin-for-all behavior.

4. Add tests proving non-admin requests are rejected at the API layer for the sensitive routes. UI tests alone won’t catch the bypass.

Once the backend enforcement is in place, this should be mergeable. Right now it creates a false sense of security: the doors disappear from the UI, but the endpoints are still reachable.

outsourc-e:
Closing during PR queue cleanup. This still has CHANGES_REQUESTED and the approach gates admin/provider settings primarily at the UI/header layer; if we revive it, it should come back as a focused PR with server-side endpoint enforcement and current Hermes naming/base branch.

PR #139, opened 2026-04-24
fix: respect chat width setting

dontcallmejames:
## Summary
- Use the configured chat content width CSS variable for message container max-width
- Apply the same chat width variable to the desktop composer wrapper

## Test Plan
- corepack pnpm build

Fixes the issue where selecting Full width still left the chat body/composer constrained to 768px.


outsourc-e (approved):
Looks good. Tiny scoped fix, uses the existing --chat-content-max-width variable consistently for the message container and desktop composer, and CI is green.

PR #166, opened 2026-04-25
Fix Docker terminal runtime

kvncrw:
Docker terminals can currently get into a half-connected state: the UI receives a `sessionId`, input requests can return 200, but no PTY output ever arrives.

The Docker runtime was missing the pieces the terminal backend assumes are available. Terminal sessions spawn `python3 pty-helper.py`, and the default shell is `/bin/bash`, but the production image only installed `ca-certificates`, `curl`, and `tini`. On top of that, helper spawn/stderr errors were emitted inside `terminal-sessions.ts` but dropped by `/api/terminal-stream`, which made the browser show a dead terminal instead of the actual failure.

This does a few small things together:

- install `python3` and `bash` in the production and dev workspace images
- create a real home directory for the `workspace` user and set `HOME`
- fall back to an existing cwd/shell when the requested defaults are unavailable
- forward terminal `error` events through the SSE stream
- show the actual terminal backend error in xterm instead of a generic connection message

Checks:

- `pnpm exec tsc --noEmit --pretty false`
- `pnpm exec eslint src/server/terminal-sessions.ts src/routes/api/terminal-stream.ts src/components/terminal/terminal-workspace.tsx`
- `pnpm build`
- production smoke: `node server-entry.js` on `127.0.0.1:4177`, then `/api/terminal-stream` + `/api/terminal-input` returned `echo terminal-ok` from the PTY stream

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,9 +25,10 @@ RUN pnpm build
 # ─── runtime stage ────────────────────────────────────────────────────────
 FROM node:22-slim
 RUN apt-get update && apt-get install -y --no-install-recommends \
-      ca-certificates curl tini \
+      bash ca-certificates curl python3 tini \
     && rm -rf /var/lib/apt/lists/* \
-    && groupadd -r workspace && useradd -r -g workspace -u 10010 workspace
+    && groupadd -r workspace \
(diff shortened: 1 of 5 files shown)


kvncrw:
Closing this for now. This was opened prematurely and should not be upstreamed in its current form.

PR #170, opened 2026-04-26
fix(chat): compact tool activity and persist local sessions

DomGrieco:
## Summary
- bring in the chat UI change that compacts consecutive tool activity into a cleaner assistant message presentation
- persist fallback local sessions under a writable Hermes home path instead of the container app directory
- preserve tool-call metadata when rehydrating local history so refreshed chats keep the expected tool activity UI

## Test plan
- [x] `corepack pnpm exec vitest run src/screens/chat/components/message-item.test.ts src/server/__tests__/local-session-store.test.ts`
- [x] `corepack pnpm exec tsc --noEmit`

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/routes/api/history.ts
+++ b/src/routes/api/history.ts
@@ -10,7 +10,73 @@ import {
 } from '../../server/hermes-api'
 import { resolveSessionKey } from '../../server/session-utils'
 import { isAuthenticated } from '@/server/auth-middleware'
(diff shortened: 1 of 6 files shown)


DomGrieco:
Closing this because it was opened against the upstream repo by mistake. The intended PR is against the fork main branch instead.

PR #191, opened 2026-04-30
fix(auth): warn at startup when Secure cookies will break plain-HTTP LAN login (#149)

### LINKED ISSUE #149: Document COOKIE_SECURE=false requirement for HOST=0.0.0.0 over plain HTTP
## Summary

Setting up the workspace per the v2.0.0 security guard — \`HOST=0.0.0.0\` with \`HERMES_PASSWORD\` over plain HTTP on a LAN — silently breaks login until the operator also sets \`COOKIE_SECURE=false\`. The error path is invisible (server returns 200, browser just discards the cookie) and there's nothing in docs or in the bootup output that mentions this.

## What happens

In \`src/server/auth-middleware.ts\`, \`shouldSetSecureCookie()\` returns \`true\` when \`NODE_ENV=production\` (the default for the systemd / docker setups), so the session cookie is set with the \`Secure\` attribute:

\`\`\`
set-cookie: hermes-auth=...; HttpOnly; Secure; SameSite=Strict; Path=/; Max-Age=2592000
\`\`\`

Browsers refuse to store \`Secure\` cookies received over plain HTTP. So:

1. \`POST /api/auth\` returns \`{\"ok\": true}\`.
2. Browser silently drops the cookie.
3. Page reloads → \`/api/auth-check\` returns \`{\"authenticated\": false}\`.
4. User is bounced back to the login screen with no useful error and no log line on the server.

The \`COOKIE_SECURE\` env var (already implemented and respected) is the correct workaround, but no part of the docs or the boot guard mentions it.

## Why this is paired with the new HOST=0.0.0.0 + HERMES_PASSWORD path

The whole reason a user is hitting this configuration is that they followed the \`[workspace] refusing to start\` error from \`server-entry.js\` (introduced in #122):

\`\`\`
HOST is set to \"0.0.0.0\" (non-loopback), but HERMES_PASSWORD is unset.
This would expose a high-privilege control plane (terminals, files, agents)
to anyone who can reach the port. Either:
  • set HOST=127.0.0.1 for local-only access, or
  • set HERMES_PASSWORD=<strong-secret> to enable workspace auth, or
  • set HERMES_ALLOW_INSECURE_REMOTE=1 to bypass this check (not recommended).
\`\`\`

A reasonable LAN deployment (home server reached over HTTP from another machine on the same subnet) goes \`HOST=0.0.0.0\` + \`HERMES_PASSWORD\`. That setup is now blessed by the boot guard, but the cookie defaults make it non-functional.

## Suggested fixes (any of these would have caught my case)

Pick whichever fits your security stance — happy to send a PR for the one you prefer:

**Option A (smallest, docs-only):** Mention \`COOKIE_SECURE=false\` in the same boot guard message that asks the operator to set \`HERMES_PASSWORD\`. Something like:

> \"Note: if you're serving over plain HTTP (not HTTPS), also set \`COOKIE_SECURE=false\` or auth cookies will be silently dropped by the browser.\"

**Option B (auto-detect):** In \`shouldSetSecureCookie()\`, treat \`HOST=0.0.0.0\` + missing TLS reverse proxy as a strong hint and default to \`false\` with a warning, unless \`COOKIE_SECURE=true\` is set explicitly. Operators on HTTPS via reverse proxy can keep the explicit \`COOKIE_SECURE=true\`. This trades zero misconfiguration cost for a small \"footgun if a user later puts the workspace behind a reverse proxy without re-reading config\".

**Option C (best UX, more work):** Detect on the *request* side that the connection is plain HTTP and emit the cookie without \`Secure\`, regardless of \`NODE_ENV\`. This is what e.g. \`express-session\` does with \`secure: 'auto'\`. Trade-off: more conditional code and slightly less predictable behavior for operators who flip TLS mid-session.

## Repro

1. Set \`HOST=0.0.0.0\`, \`HERMES_PASSWORD=...\` in \`.env\`.
2. Do **not** set \`COOKIE_SECURE\`.
3. Run with \`NODE_ENV=production\` (default for the systemd template).
4. Browse to \`http://<lan-ip>:3333/\` from another machine, enter password.
5. Stuck on the login screen. \`document.cookie\` is empty. Server logs show the POST returned 200.

## My current workaround

Added \`COOKIE_SECURE=false\` to the workspace \`.env\`. Took a debug session to find that env var — it doesn't appear in \`.env.example\`, in the boot guard message, in the README, or in the docs site. Hopefully this issue helps the next person spend less time on it.

Interstellar-code:
## Summary

- Adds a startup warning in `server-entry.js` when `HOST` is non-loopback, `NODE_ENV=production`, and `COOKIE_SECURE` is not explicitly set to `0`/`false`/`no` — the exact conditions that cause browser-silent login failure over plain HTTP
- Documents `COOKIE_SECURE=0` in `.env.example` alongside the existing `=1` case
- Adds `COOKIE_SECURE` entry to the README env-vars table

## Root cause

`shouldSetSecureCookie()` in `auth-middleware.ts` returns `true` under `NODE_ENV=production`, setting the `Secure` attribute on session cookies. Browsers silently drop `Secure` cookies over `http://`, so `POST /api/auth` returns `{"ok":true}` but the cookie is never stored — login loops with no error.

The `COOKIE_SECURE` env var already exists and fixes it, but nothing told the operator it was needed.

## No logic changes

`auth-middleware.ts` is untouched — it already handles `COOKIE_SECURE=0` correctly. This PR is purely operator-facing: a runtime warning and two documentation additions.

## Test plan

- [ ] Set `HOST=0.0.0.0`, `HERMES_PASSWORD=x`, `NODE_ENV=production`, no `COOKIE_SECURE` → confirm warning prints at startup
- [ ] Add `COOKIE_SECURE=0` → confirm warning does not print
- [ ] Loopback (`HOST=127.0.0.1`) → confirm no warning regardless of `NODE_ENV`

Fixes #149

Worked with Interstellar Code

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -76,8 +76,13 @@
 #
 # Set to 1 to force the Secure attribute on session cookies even when
 # NODE_ENV is not production — useful when terminating TLS at a reverse
-# proxy. Set to 0 to force it off (only safe for localhost HTTP).
+# proxy.
 # COOKIE_SECURE=1
+#
+# Set to 0 when running a plain-HTTP LAN deployment (HOST=0.0.0.0 without
+# HTTPS). NODE_ENV=production enables Secure cookies by default; browsers
+# silently drop Secure cookies over http://, causing login to silently fail.
+# COOKIE_SECURE=0
 
 # Trust proxy-forwarded headers (default: off)
 #
--- a/README.md
+++ b/README.md
@@ -547,6 +547,7 @@ Features pending cloud infrastructure:
 
 - `HERMES_PASSWORD` — required whenever `HOST ≠ 127.0.0.1`
 - `COOKIE_SECURE=1` — force the `Secure` cookie flag when terminating HTTPS at a proxy
+- `COOKIE_SECURE=0` — disable the `Secure` flag for plain-HTTP LAN deployments (`HOST=0.0.0.0` without HTTPS); without this, browsers silently drop session cookies and login fails (#149)
 - `TRUST_PROXY=1` — trust `x-forwarded-for` / `x-real-ip` (only set behind a sanitizing reverse proxy)
 - `HERMES_DASHBOARD_TOKEN` — explicit bearer for dashboard API (preferred over the legacy HTML-scrape fallback)
 - `HERMES_ALLOW_INSECURE_REMOTE=1` — bypass the fail-closed guard (not recommended)
--- a/server-entry.js
+++ b/server-entry.js
@@ -45,6 +45,21 @@ if (isNonLoopbackHost(host)) {
       '[workspace] HERMES_ALLOW_INSECURE_REMOTE is set — starting anyway.',
     )
   }
+
+  // Warn when serving over plain HTTP with a password: NODE_ENV=production
+  // sets the Secure flag on session cookies, which browsers silently drop
+  // over http://.  Operators must set COOKIE_SECURE=0 for plain-HTTP LAN
+  // deployments.  See #149.
+  const cookieSecureOverride = (process.env.COOKIE_SECURE || '').trim().toLowerCase()
+  const cookieSecureExplicit = cookieSecureOverride === '0' || cookieSecureOverride === 'false' || cookieSecureOverride === 'no'
+  if (!cookieSecureExplicit && process.env.NODE_ENV === 'production') {
+    console.warn(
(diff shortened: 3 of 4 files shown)


outsourc-e:
Cherry-picked the focused fix from this PR onto main and pushed it in the v2.1.3 cleanup train. Closing this stacked PR to avoid merging the older branch state.

PR #194, opened 2026-04-30
fix(models): always pin config default model first in /api/models response

### LINKED ISSUE #193: Model selector always shows OpenRouter model instead of config default
## Bug

The model button in the chat composer always displays `anthropic/claude-sonnet-4-20250514` (or whatever the first `models.json` entry is) instead of the model configured as default in `~/.hermes/config.yaml`.

## Root cause

`/api/models` reads `~/.hermes/models.json` in order, then calls `readHermesDefaultModel()` to get the config.yaml default. It inserts the default only **when it is absent** from the list:

```ts
const hasDefault = models.some((m) => m.id === defaultModel.id)
if (!hasDefault) {
  models.unshift(defaultModel)
}
```

OpenRouter entries store their model ID in the format `anthropic/claude-sonnet-4-20250514` (provider-prefixed). When this is the first entry in `models.json`, it becomes `models[0]`.

`ChatComposer` uses `configuredModel = models[0].id` as the label fallback for new chats (before a session exists). Because the config default (`glm-5.1`) **is** already in the list at position 4, it is never moved — position 0 remains the OpenRouter entry, so the wrong model is shown.

## Steps to reproduce

1. Have `models.json` with an OpenRouter entry as the first record (model field `"anthropic/claude-sonnet-4-20250514"`)
2. Have a different model set as default in `config.yaml` (e.g. `model.default: glm-5.1`)
3. Open a new chat — model selector shows `anthropic/claude-sonnet-4-20250514`

## Fix

Always move the config default to position 0, replacing any existing duplicate entry:

```ts
const defaultModel = readHermesDefaultModel()
if (defaultModel) {
  models = models.filter((m) => m.id !== defaultModel.id)
  models.unshift(defaultModel)
}
```

Worked with Interstellar Code

Interstellar-code:
Fixes #193

## Summary

- `/api/models` previously only inserted the `config.yaml` default model when it was absent from `models.json` — if it was already present at position 4, it stayed there
- OpenRouter entries use provider-prefixed model IDs (`anthropic/claude-sonnet-4-20250514`), so the first `models.json` entry became `models[0]`
- `ChatComposer` uses `models[0].id` as the label fallback for new chats, causing it to always display the OpenRouter entry instead of the configured default
- Fix: always filter-out then unshift the config default so it is unconditionally first

## Test plan

- [ ] Open a new chat with an OpenRouter entry as first in `models.json` and a different model set as default in `config.yaml` — model selector should show the config default
- [ ] Existing sessions should continue showing their session model unchanged
- [ ] Model switching via the selector should still work normally

Worked with Interstellar Code


outsourc-e:
Cherry-picked the focused fix from this PR onto main and pushed it in the v2.1.3 cleanup train. Closing this stacked PR to avoid merging the older branch state.

PR #196, opened 2026-04-30
fix(chat): make long-running SSE streams survive silent agent processing

### LINKED ISSUE #195: Chat SSE streams disconnect during long silent agent runs (multiple root causes)
## Summary

When an upstream LLM provider produces output slowly or silently for more than ~2 minutes, the workspace chat UI tears down the SSE stream prematurely even though the agent is still alive on the gateway side. After investigating one such session in depth, we identified **five distinct root causes** that all contribute to the same symptom — the gateway logs `SSE client disconnected; interrupted agent task` and the chat UI shows a stalled or empty response.

## How we hit it

Tested with the Z.AI coding plan (`https://api.z.ai/api/coding/paas/v4`, `glm-5.1`). Under load the provider:

- Returns 429 rate limits (very low concurrency on the coding plan — `glm-5.1` is 1 concurrent)
- Drops connections mid-stream (`RemoteProtocolError: peer closed connection without sending complete message body`)
- Stops sending bytes for 60–180s during heavy reasoning, then resumes

During those silent windows the gateway is happily waiting on Z.AI but the workspace was killing the SSE early. Switching to a different provider made the symptom go away — confirming the workspace had a real bug, not just a slow upstream.

We saw multiple sessions die at suspiciously specific durations:

| Session | Duration | Gateway bytes | Likely cause |
|---|---|---|---|
| A | 112s | 0 | client poller false-positive |
| B | 136s | 0 | cross-session abort + 120s timer |
| C | 315s | 0 | Node.js default 5-minute `requestTimeout` |
| D | various | 0 | `AbortError` left UI flags stuck |

### 1. No application-layer heartbeat on `send-stream.ts`
`src/routes/api/send-stream.ts` did not emit periodic events during silent gateway processing. The client-side no-activity timer in `use-streaming-message.ts` (\`ACCEPTED_NO_ACTIVITY_TIMEOUT_MS\` = 120s) fired and aborted otherwise-healthy streams.

### 2. Active-run poller false-positives
The fallback poller in \`chat-screen.tsx\` (lines ~901-920) fetched \`/api/sessions/:key/active-run\` every 5s and called \`streamFinish()\` whenever \`!data.run\` was truthy or \`status\` was outside \`['accepted','active','handoff']\`. Three failure modes:

- Gateway hadn't registered the run yet during silent processing → \`data.run\` was null → \`streamFinish()\` fires while stream is still alive
- Transient statuses (\`queued\`, \`pending\`) treated as terminal
- Polled even when send-stream was actively delivering chunks (the \`sseConnectionState\` it gates on is the *separate* chat-events SSE channel)

### 3. Node.js default 5-minute `requestTimeout` (Node 18+)
The Vite dev server has no override for \`server.httpServer.requestTimeout\`. Node's default 300_000ms cap killed any HTTP request older than 5 minutes — including long-running SSE responses. Session C's death at exactly 315s pointed straight at this.

### 4. `AbortError` path left UI flags stuck
\`use-streaming-message.ts\` \`startStreaming\` caught \`AbortError\` and \`return\`-ed silently. \`chat-screen.tsx\` continued holding \`sending=true\` / \`waitingForResponse=true\`, blocking the composer from accepting another message. Triggered when:

- User sends a message in session B while session A is streaming (line 716 unconditional abort — see #5 below)
- HMR-induced unmount/remount during dev
- Vite dev server auto-restart

### 5. Hardcoded SSE timeouts not configurable
\`ACCEPTED_NO_ACTIVITY_TIMEOUT_MS\` and \`HANDOFF_NO_ACTIVITY_TIMEOUT_MS\` were hardcoded constants (120s / 180s). Slow providers + heavy contexts need much longer windows. Defaults should be sourced from \`~/.hermes/config.yaml\` (\`workspace.stream_accepted_timeout\` / \`workspace.stream_handoff_timeout\`) so users can tune without rebuilding.

## Additional issue surfaced (not fixed in this report)

\`use-streaming-message.ts:716\` unconditionally aborts \`eventSourceRef.current\` whenever \`startStreaming\` is called for a new message. The hook is a singleton per chat-screen instance, so sending a message to session B kills any in-flight stream from session A. This is incompatible with concurrent multi-session chats. Proper fix requires keying state per \`sessionKey\` (\`Map<sessionKey, AbortController>\`) — non-trivial refactor of the hook.

## Proposed fixes (PR will follow)

1. Add 30s heartbeat to \`send-stream.ts\` (\`event: heartbeat\`, before the try block, inside the ReadableStream \`start()\` callback)
2. Add \`case 'heartbeat': markActivity(); break;\` to the SSE event switch in \`use-streaming-message.ts\`
3. Read \`workspace.stream_accepted_timeout\` / \`workspace.stream_handoff_timeout\` from \`~/.hermes/config.yaml\`, expose via \`/api/models\` response, consume in \`chat-screen.tsx\`
4. Tighten active-run poller — only fire \`streamFinish()\` on definitively terminal statuses (\`completed|failed|cancelled|error\`); treat null/queued/pending as keep-waiting
5. Disable Node \`requestTimeout\` for the Vite dev server in \`vite.config.ts\` (\`server.httpServer.requestTimeout = 0\`)
6. Add \`onAbort\` callback option to \`useStreamingMessage\`; in the \`AbortError\` branch, clear hook state and call \`onAbort\` so \`chat-screen.tsx\` can reset \`sending\` / \`waitingForResponse\` / \`pendingGeneration\` flags

## Result

After all six fixes, long-running agent sessions on slow providers complete normally. Verified with multi-minute reasoning runs that previously died at the 2-minute and 5-minute marks.

## Files touched

- \`src/routes/api/send-stream.ts\`
- \`src/routes/api/models.ts\`
- \`src/screens/chat/hooks/use-streaming-message.ts\`
- \`src/screens/chat/chat-screen.tsx\`
- \`vite.config.ts\`

Interstellar-code:
Closes #195.

## Summary

Fixes five distinct bugs that caused chat SSE streams to disconnect during multi-minute silent processing windows on slow upstream providers. Discovered while testing against the Z.AI coding plan (`glm-5.1` via `https://api.z.ai/api/coding/paas/v4`) — that plan has 1 concurrent request and frequent connection drops, which forced the gateway into silent windows of 60–180 seconds at a time. Multiple sessions died at suspiciously specific durations (~112s, ~136s, exactly 315s) — each pointing at a different layer.

See issue #195 for the full root cause walkthrough.

## What this PR changes

**1. Server-side heartbeat (`src/routes/api/send-stream.ts`)**
30-second `event: heartbeat` ticks emitted from the SSE stream so the client-side no-activity timer doesn't fire during silent gateway processing.

**2. Client heartbeat handler (`src/screens/chat/hooks/use-streaming-message.ts`)**
`case 'heartbeat': markActivity(); break;` added to the SSE event switch. Resets the no-activity timer and transitions the lifecycle phase out of `accepted` once any signal arrives.

**3. Configurable timeouts (`src/routes/api/models.ts` + `src/screens/chat/chat-screen.tsx`)**
`/api/models` now exposes `streamAcceptedTimeoutMs` and `streamHandoffTimeoutMs`, sourced from `workspace.stream_accepted_timeout` / `workspace.stream_handoff_timeout` in `~/.hermes/config.yaml`. Env vars `STREAM_ACCEPTED_TIMEOUT_MS` / `STREAM_HANDOFF_TIMEOUT_MS` still override. `chat-screen.tsx` feeds them into `useStreamingMessage` so users can tune without rebuilding. `.env.example` documents the env var path.

**4. Active-run poller false-positive (`src/screens/chat/chat-screen.tsx`)**
The fallback poller used to call `streamFinish()` whenever `data.run` was null or status was outside `['accepted','active','handoff']`. Both conditions hit during normal silent processing — the gateway hadn't registered the run yet, or the status was a transient `queued`/`pending`. Tightened to fire only on definitively terminal statuses (`completed|failed|cancelled|error`), treats null run / unknown status as keep-waiting.

**5. Node.js requestTimeout disabled in dev (`vite.config.ts`)**
Node 18+ defaults `server.httpServer.requestTimeout` to 5 minutes, which kills any HTTP request older than that — including healthy long-running SSE streams. Set to 0 for `command === 'serve'` only. Production servers keep their defaults to avoid slowloris exposure.

**6. AbortError cleanup (`src/screens/chat/hooks/use-streaming-message.ts` + `src/screens/chat/chat-screen.tsx`)**
`startStreaming` previously caught `AbortError` and returned silently, leaving `chat-screen.tsx` holding `sending=true` / `waitingForResponse=true` / `pendingGeneration=true`. The composer would block subsequent submits indefinitely. Added an optional `onAbort` callback option; `AbortError` branch now clears hook state and invokes `onAbort()` so callers can reset their flags. `chat-screen.tsx` wires this to clear all three flags.

## Commits

- `0a8b633` — server-side 30s heartbeat
- `c8d4572` — configurable timeouts via config.yaml + /api/models exposure
- `941f8ef` — poller fix, dev-only requestTimeout=0, AbortError cleanup, model.ts timeout reading

## Known follow-up (not in this PR)

`use-streaming-message.ts` is a singleton hook per chat-screen instance. Sending a message in session B unconditionally aborts any in-flight stream from session A (line ~716). Incompatible with concurrent multi-session chat. Proper fix requires keying state per `sessionKey` (`Map<sessionKey, AbortController>`) — a non-trivial refactor outside the scope of this PR. Captured in issue #195.

## Test plan

- [x] Long-running session (>2 min) on a slow provider (Z.AI `glm-5.1` with rate limits) completes without disconnect — verified after Z.AI was swapped for a faster fallback
- [x] Long-running session (>5 min) doesn't hit Node's 5-minute request timeout
- [x] Cross-session navigation no longer leaves composer stuck (sending/waitingForResponse cleared)
- [x] `tsc --noEmit` passes
- [ ] Reviewer to verify against their own slow-provider scenario


outsourc-e:

PR #198, opened 2026-04-30
fix(sessions): allow deleting local-only sessions in portable mode

### LINKED ISSUE #197: DELETE /api/sessions silently fails for local-only sessions in portable mode
## Summary

The DELETE handler in \`src/routes/api/sessions.ts\` doesn't account for sessions persisted in the workspace local store (\`.runtime/local-sessions.json\`). In portable-mode setups (no enhanced sessions API on the gateway), DELETE requests either:

- Silently return \`{ deleted: false }\` when \`capabilities.sessions === false\` — the UI optimistically removes the session but it reappears on reload because the workspace local store still has it.
- Return HTTP 500 when \`capabilities.sessions === true\` but the gateway's \`/api/sessions/{id}\` endpoint actually 404s for the local-only session ID. Browser sees \`Hermes dashboard /api/sessions/api-xxxx: 404 {\"detail\":\"Session not found\"}\`.

Either way, the delete button in the UI doesn't work for sessions that were created in portable mode.

## How we hit it

Running with \`HERMES_API_URL=http://127.0.0.1:8642\` against a hermes-agent gateway. Workspace probe reported \`mode=zero-fork core=[..., chatCompletions, ..., streaming, dashboard] enhanced=[sessions, ...] missing=[enhancedChat]\` so \`capabilities.sessions === true\`. But the actual gateway endpoint \`/api/sessions/{id}\` returns 404 — there's a mismatch between the capability probe result and what the API actually serves for portable-only sessions.

The session was a portable session created by \`send-stream.ts\` (which calls \`ensureLocalSession()\` in \`src/server/local-session-store.ts\`). It lives only in \`.runtime/local-sessions.json\`. The gateway has no record of it. DELETE proxied to the gateway, got 404, returned 500.

## Reproduction

1. Run workspace + hermes-agent in portable / zero-fork mode (no enhanced sessions API)
2. Start a new chat — \`api-XXXXX\` session is created in the local store
3. Click delete on the session in the UI
4. Observe HTTP 500 with \`Hermes dashboard /api/sessions/api-XXXX: 404 ...\` (or silent no-op + session reappearing on reload, depending on capability flag)

## Proposed fix

In \`DELETE /api/sessions\`, check \`getLocalSession(sessionKey)\` from \`src/server/local-session-store.ts\` *before* the capability check. If the session lives in the local store, call \`deleteLocalSession()\` and return \`{ ok: true, source: 'local' }\`. Only fall through to the gateway delete path when the session is not in the local store.

PR follows.

## Files

- \`src/routes/api/sessions.ts\` (DELETE handler)
- \`src/server/local-session-store.ts\` (provides \`getLocalSession\` + \`deleteLocalSession\`, already exported)

Interstellar-code:
Closes #197.

## Summary

DELETE \`/api/sessions\` now checks the workspace local session store before proxying to the gateway, fixing the case where portable-mode sessions (persisted only in \`.runtime/local-sessions.json\`) couldn't be deleted via the UI.

Before this PR:
- \`capabilities.sessions === false\` path: returned \`{ deleted: false }\` silently — UI removes session optimistically, it reappears on reload
- \`capabilities.sessions === true\` path: proxied to gateway, got 404 for local-only sessions, returned HTTP 500 to browser

After:
- Local-only sessions: handled by \`deleteLocalSession()\`, returns \`{ ok: true, source: 'local' }\`
- Gateway-tracked sessions: unchanged path

## How it was found

Running workspace against a hermes-agent gateway in zero-fork mode where the capability probe reported sessions API was available, but \`GET /api/sessions/{id}\` actually 404s for portable-only sessions. The mismatch caused the delete button in the UI to throw \`Hermes dashboard /api/sessions/api-XXXX: 404 {\"detail\":\"Session not found\"}\` for any chat created via the portable code path.

See #197 for full root-cause walkthrough.

## Test plan

- [x] DELETE on local-only session returns 200 with \`source: 'local'\`
- [x] Session disappears from \`.runtime/local-sessions.json\`
- [x] Gateway-tracked session DELETE path unchanged
- [x] \`tsc --noEmit\` passes

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -77,7 +77,8 @@
 # Set to 1 to force the Secure attribute on session cookies even when
 # NODE_ENV is not production — useful when terminating TLS at a reverse
 # proxy. Set to 0 to force it off (only safe for localhost HTTP).
-# COOKIE_SECURE=1
+# COOKIE_SECURE=1   # use behind HTTPS / reverse proxy
+# COOKIE_SECURE=0   # required for plain HTTP on HOST=0.0.0.0 / LAN access
 
 # Trust proxy-forwarded headers (default: off)
 #
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,7 +25,7 @@ RUN pnpm build
 # ─── runtime stage ────────────────────────────────────────────────────────
 FROM node:22-slim
 RUN apt-get update && apt-get install -y --no-install-recommends \
-      ca-certificates curl tini \
+      ca-certificates curl tini python3 \
     && rm -rf /var/lib/apt/lists/* \
     && groupadd -r workspace && useradd -r -g workspace -u 10010 workspace
 
--- a/README.md
+++ b/README.md
@@ -547,6 +547,7 @@ Features pending cloud infrastructure:
 
 - `HERMES_PASSWORD` — required whenever `HOST ≠ 127.0.0.1`
 - `COOKIE_SECURE=1` — force the `Secure` cookie flag when terminating HTTPS at a proxy
+- `COOKIE_SECURE=0` — required for plain HTTP LAN access on `HOST=0.0.0.0` so the browser will actually send the auth cookie
 - `TRUST_PROXY=1` — trust `x-forwarded-for` / `x-real-ip` (only set behind a sanitizing reverse proxy)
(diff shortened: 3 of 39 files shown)


outsourc-e:

PR #202, opened 2026-04-30
fix(api): add dashboard fallback for createSession, updateSession, forkSession

Armanoide:
## Summary
`createSession`, `updateSession`, and `forkSession` were missing the dashboard availability check that all other session functions already have (`listSessions`, `getSession`, `deleteSession`, `getMessages`, `searchSessions`). When the dashboard is available, requests now route through it instead of hitting the gateway API directly.

## Root cause
- `listSessions`, `getSession`, `deleteSession`, `getMessages`, `searchSessions` all check `getCapabilities().dashboard.available` and use the dashboard API when it's reachable
- `createSession`, `updateSession`, `forkSession` had no such check — they always called the gateway directly, bypassing the dashboard even when it was available
- This meant session mutations couldn't leverage the dashboard's session management, causing inconsistency in deployments where the dashboard is the primary backend

## Fix
Added the three missing functions to hermes-dashboard-api.ts and wired them into the existing dashboard fallback pattern in `hermes-api.ts`:
```
// hermes-dashboard-api.ts — new functions
export async function createSession(body) → POST /api/sessions
export async function updateSession(id, body) → PATCH /api/sessions/:id
export async function forkSession(id) → POST /api/sessions/:id/fork
```
```
// hermes-api.ts — added dashboard guard (same pattern as getSession)
if (getCapabilities().dashboard.available) {
  return dashboardFn(...) as Promise<HermesSession>
}
// fall through to direct gateway API
```
No behavior change when the dashboard is not available — the gateway API remains the fallback.

## Test plan
- [ ] Dashboard available: session creation/update/fork routes through dashboard API
- [ ] Dashboard unavailable: falls through to gateway API — no regression
- [ ] Type check: as HermesSession casts follow the same pattern as getSession and listSessions

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -77,7 +77,8 @@
 # Set to 1 to force the Secure attribute on session cookies even when
 # NODE_ENV is not production — useful when terminating TLS at a reverse
 # proxy. Set to 0 to force it off (only safe for localhost HTTP).
-# COOKIE_SECURE=1
+# COOKIE_SECURE=1   # use behind HTTPS / reverse proxy
+# COOKIE_SECURE=0   # required for plain HTTP on HOST=0.0.0.0 / LAN access
 
 # Trust proxy-forwarded headers (default: off)
 #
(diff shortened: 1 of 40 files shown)


outsourc-e:
Cherry-picked the focused fixes from this stacked PR onto current main, resolved conflicts against Hermes naming/home changes, verified targeted tests + production build, and moved the v2.1.3 tag forward. Closing the stale stacked branch to keep the queue clean.

PR #210, opened 2026-05-01
fix: keep chat sends recoverable after navigation

dterenyi:
## Summary
- Keep the upstream Hermes chat run alive when the Chat view unmounts.
- Persist active run state in `run-store` so active-run polling can resume after navigation.
- Recover pending optimistic sends after session resolution changes the canonical session key.

## Test plan
- `pnpm build`
- `git diff --check`

## Notes
- Full `tsc` and focused lint are still blocked by pre-existing workspace baseline issues outside this fix.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -77,7 +77,8 @@
 # Set to 1 to force the Secure attribute on session cookies even when
 # NODE_ENV is not production — useful when terminating TLS at a reverse
 # proxy. Set to 0 to force it off (only safe for localhost HTTP).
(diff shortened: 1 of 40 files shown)


outsourc-e:
Cherry-picked the focused fixes from this stacked PR onto current main, resolved conflicts against Hermes naming/home changes, verified targeted tests + production build, and moved the v2.1.3 tag forward. Closing the stale stacked branch to keep the queue clean.

PR #212, opened 2026-05-01
Fix conductor mission tracking and portable fallback

dterenyi:
Fixes two conductor issues:\n- Portable missions now stream through /api/send-stream instead of the dead scheduled-jobs path.\n- Dashboard-backed missions are matched by recent mission text as well as exact keys, so the conductor UI shows live activity instead of 0 active.\n\nValidation:\n- pnpm run build\n- Browser reload of /conductor showed 1 active and the spawned worker card.\n\nBug log: docs/conductor-bug-log.md

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -77,7 +77,8 @@
 # Set to 1 to force the Secure attribute on session cookies even when
 # NODE_ENV is not production — useful when terminating TLS at a reverse
 # proxy. Set to 0 to force it off (only safe for localhost HTTP).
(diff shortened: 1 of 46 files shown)


outsourc-e:
Cherry-picked the conductor mission tracking + portable fallback fix onto current main, resolved conflicts against the current Hermes codebase, verified targeted tests + production build, and pushed it in the v2.1.3 cleanup train. Closing this stale stacked PR.

PR #213, opened 2026-05-01
fix: load workspace env in stable launcher

tumbleweedlabs:
## Summary
- Load the Workspace .env file before deriving runtime settings in scripts/start-stable.sh
- Ensures non-interactive shells/services get Hermes API/dashboard URLs and tokens during build/startup
- Avoids misleading auth/provider failures when the stable launcher starts without exported env vars

## Test Plan
- bash -n scripts/start-stable.sh
- git diff --check
- pnpm test
- pnpm build


outsourc-e:

PR #226, opened 2026-05-02
Treat missing worker chat DB as no-session state

Interstellar-code:
## Summary
- Treat a missing worker `state.db` as an unavailable/no-session state instead of a live-chat error
- Add a regression test for uninitialized worker profiles

## Verification
- `pnpm vitest run src/server/swarm-chat-reader.test.ts`
- `pnpm build`


outsourc-e:
Closing as duplicate of #232 (identical diff, same author). Merged via #232 — thanks @Interstellar-code 🙏

PR #288, opened 2026-05-03
fix(dev): auth header on gateway probes + expanded vite watch.ignored

Interstellar-code:
## Summary

- Sends `Authorization: Bearer <HERMES_API_TOKEN>` on startup connection-status probes (`/v1/models`, `/api/sessions`) so the dev server correctly detects a gateway secured with `API_SERVER_KEY`
- Adds comprehensive `watch.ignored` list to prevent spurious HMR reloads from:
  - `routeTree.gen.ts` (TanStack Router regenerates this on every route change → infinite loop without the ignore)
  - `.runtime/**` (session store rewritten on every chat message)
  - `.tanstack/**`, `.omc/**`, `.omx/**`, `.claude/**`, `.cursor/**`, `.vscode/**`
  - Build artifacts, test outputs, log files

## Test plan

- [ ] `pnpm dev` with `API_SERVER_KEY` set — gateway capability probe returns 200 instead of 401
- [ ] Edit a route file — HMR fires once, not in an infinite loop
- [ ] Chat send — no spurious server reload from `.runtime/` write

Worked with Interstellar Code

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -474,6 +474,36 @@ const config = defineConfig(({ mode, command }) => {
       port: process.env.PORT ? Number(process.env.PORT) : 3000,
       strictPort: false, // allow fallback if port is taken, but log clearly
       allowedHosts: true,
+      watch: {
+        ignored: [
+          // Generated route tree — TanStack Router's file watcher detects its
+          // own output as a change → infinite regeneration loop.
+          '**/routeTree.gen.ts',
+          // Local portable session store, rewritten on every chat send.
+          // Without this, the watcher fires on every message → spurious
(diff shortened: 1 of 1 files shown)


outsourc-e:
Thanks @Interstellar-code! Landed the auth header on probes + the watch.ignored list via integrated commit ce4fab300 on main. One change from your patch: dropped '**/routeTree.gen.ts' from the ignore list because there's an existing regression test (src/router-route-resolution.test.ts) that forbids it — ignoring it breaks route HMR. The real root cause of the gen-tree thrash was multiple concurrent vite dev servers writing to the same file, which was fixed separately today. Closing since the rest of the change is merged.

PR #289, opened 2026-05-03
fix(vite): probe /api/sessions on dashboard URL and cache result

### LINKED ISSUE #276: Vite dev server polls /api/sessions on wrong port, causing 404 errors
The connection-status handler in vite.config.ts makes requests to a hardcoded http://localhost:8642 (the agent/gateway port). However, the /api/sessions endpoint only exists on the dashboard port (9119 by default), not on the agent port.

This causes a 404 error every 15 seconds in the browser console and server logs when running in a setup where the dashboard is on a separate port.

Steps to reproduce:

Run Hermes Agent on port 8642
Run Hermes Dashboard on port 9119
Start Hermes Workspace dev server
Open browser console — 404 errors for http://localhost:8642/api/sessions?limit=1 appear every 15 seconds
Expected behavior:
The workspace should either:

Query /api/sessions on the configured dashboard URL (CLAUDE_DASHBOARD_URL / HERMES_DASHBOARD_URL), or
Cache the "portable mode" detection to avoid repeated 404 requests
Actual behavior:
vite.config.ts:533 hardcodes const agentUrl = 'http://localhost:8642' and polls /api/sessions there, which always returns 404.

Environment:

Hermes Workspace: latest main
Hermes Agent: running on 8642
Hermes Dashboard: running on 9119
CLAUDE_DASHBOARD_URL / HERMES_DASHBOARD_URL correctly set to http://127.0.0.1:9119
Additional context:
The API route /api/connection-status in the workspace already correctly uses ensureGatewayProbed() from gateway-capabilities.ts, which properly checks the dashboard URL. Only the Vite dev server's inline handler has this issue.

Kind regards

Sanjays2402:
## Summary

Fixes #276.

The Vite dev server's inline `/api/connection-status` handler in `vite.config.ts` probed `/api/sessions` on the **agent** URL (default `http://127.0.0.1:8642`). The Hermes Agent does not serve `/api/sessions` — the Hermes Dashboard does (default `http://127.0.0.1:9119`). Anywhere the dashboard runs on a separate port (any setup that sets `CLAUDE_DASHBOARD_URL` / `HERMES_DASHBOARD_URL`), this produced a 404 in the browser console every 15 seconds.

The reporter notes that the production-side route `src/server/gateway-capabilities.ts` already does the right thing via `ensureGatewayProbed()`; only the dev-server inline handler had the bug.

## Fix

Two small changes in `vite.config.ts`:

1. **Resolve a dedicated dashboard URL** from env (`HERMES_DASHBOARD_URL`, `CLAUDE_DASHBOARD_URL`, fallback `http://127.0.0.1:9119`) and probe `/api/sessions?limit=1` there. This mirrors the resolution path already used by `gateway-capabilities.ts`.
2. **Cache the resolved payload** for 60s on success (5s on failure) so a tight poll loop doesn't fan out three fetches every cycle when the answer hasn't changed.

The `/v1/models` and `/health` probes still hit the agent URL — portable-mode detection is unchanged.

## Behaviour

Before:
- Agent on `:8642`, dashboard on `:9119` → `GET http://localhost:8642/api/sessions?limit=1` → 404 every 15s.

After:
- Same setup → `GET http://localhost:9119/api/sessions?limit=1` → 200 once, then served from cache for 60s.
- Single-port setups (agent and dashboard collocated, or dashboard absent) keep working: `hasSessions` falls to `false`, falls through to portable mode, then `/health` — same as before.

## Verification

```
pnpm dev
# in another shell:
curl http://localhost:3000/api/connection-status
# {"ok":true,"mode":"portable","backend":"http://127.0.0.1:8642"}
curl http://localhost:3000/api/connection-status
curl http://localhost:3000/api/connection-status
# all three return immediately, only the first triggers outbound probes
```

`pnpm test` — 547 passing, 11 failing. All 11 failures are **pre-existing on `main`** (markdown plugin exports, chat-message-list, send-stream `persistActiveRun`, swarm-kanban types) and have nothing to do with this change. Verified by reverting the patch and re-running: identical failure set.

`tsc --noEmit` reports no errors in `vite.config.ts`. `prettier --check` passes. `vite.config.ts` is in the eslint ignore set so eslint is not applicable.

## Scope

Single file. No public API change. No env var rename. Production code path (`src/server/gateway-capabilities.ts`) is untouched.

Fixes outsourc-e/hermes-workspace#276

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -34,10 +34,10 @@ function resolveClaudeAgentDir(env: Record<string, string>): string | null {
   // Resolve relative to the workspace root (parent of hermes-workspace/)
   const workspaceRoot = dirname(resolve('.'))
   candidates.push(
-    resolve(workspaceRoot, 'hermes-agent'),            // sibling (old README)
-    resolve(workspaceRoot, '..', 'hermes-agent'),      // one level up
-    resolve(os.homedir(), '.claude', 'hermes-agent'),  // Nous installer default
-    resolve(os.homedir(), 'hermes-agent'),             // ~/hermes-agent
+    resolve(workspaceRoot, 'hermes-agent'), // sibling (old README)
+    resolve(workspaceRoot, '..', 'hermes-agent'), // one level up
+    resolve(os.homedir(), '.claude', 'hermes-agent'), // Nous installer default
+    resolve(os.homedir(), 'hermes-agent'), // ~/hermes-agent
   )
 
   for (const candidate of candidates) {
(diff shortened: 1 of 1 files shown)


outsourc-e:
Thanks @Sanjays2402! Landed the dashboard-URL probe + cache via integrated commit ce4fab300 on main, alongside @Interstellar-code's auth-header work from #288. Closes #276 as well. Closing this PR since the change is in.

PR #299, opened 2026-05-04
fix(pwa): re-enable service worker with caching, restore install prompt

aliasfoxkde:
## Summary

Fixes the PWA install option on hermes-workspace.com that stopped working after the user uninstalled and cleared cache.

## Changes

1. **public/sw.js** — Replaces the DISABLED no-op service worker that was unregistering itself and clearing all caches on every page load. New SW uses stale-while-revalidate caching for static assets, network-only for API calls.

2. **public/manifest.json** — Explicitly sets `display: standalone` to trigger Chrome's native install mini-info bar on return visits after SW activation.

3. **src/routes/__root.tsx** — Removed the unconditional `unregisterServiceWorkers()` call that ran on every page load (this was the root cause: after uninstall, the browser tried to register the SW, the app's useEffect immediately unregistered it again, so the PWA could never reinstall). Replaced with `navigator.serviceWorker.register('/sw.js')`.

4. **viewport meta** — Removed `viewport-fit=cover` and `interactive-widget=resizes-visual` which were extending content into iOS safe areas and covering navigation buttons.

## Testing

After merge: hard-refresh hermes-workspace.com, visit in Chrome Android, see install banner. On iOS Safari, use Share > Add to Home Screen.

### DIFF
--- a/public/sw.js
+++ b/public/sw.js
@@ -1,24 +1,53 @@
-// Hermes Workspace Service Worker — DISABLED
-// Unregisters itself and clears all caches to prevent stale asset issues
-// after Docker image updates or reverse proxy deployments.
+// Hermes Workspace Service Worker
+// Hermes SW v2 — stale-while-revalidate caching, auto-update on redeploy
+const CACHE_NAME = 'hermes-ws-v2';
+const PRECACHE_ASSETS = [
+  '/',
+  '/manifest.json',
+  '/claude-icon-192.png',
+  '/claude-icon-512.png',
+  '/apple-touch-icon.png',
+];
 
-self.addEventListener('install', () => {
-  self.skipWaiting()
-})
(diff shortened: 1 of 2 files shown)


outsourc-e:
Thanks for re-enabling SW — install prompt is great. Before merge, we need either:

(A) Add the manifest-side change so install prompt actually has a complete PWA manifest (icons, theme_color, scope, start_url all matching workspace)

— OR —

(B) Scope this PR down to just the SW caching changes, and we'll do manifest changes in a follow-up PR

Right now we're in a half-state where SW is back but manifest isn't fully wired for an install-quality experience. Either path is fine, just want to land it consistently. Which do you prefer?

— triaged 2026-05-05

outsourc-e:

PR #308, opened 2026-05-04
fix: allow workspace production server to start

jarnodevries-byte:
## Summary
- Remove a duplicate `cookieSecureOverride` declaration in `server-entry.js`.
- This lets the static/production Workspace server start cleanly instead of throwing a syntax error.

## Why
Jarno's local launchd Workspace was running the Vite dev server, which caused periodic HMR/page reloads in the UI. Switching launchd to the static production server exposed this duplicate declaration. Removing it makes production launchd viable and avoids the recurring dev-server refresh loop.

## Verification
- `node --check server-entry.js`
- Local launchd switched to `NODE_ENV=production` + `pnpm exec node server-entry.js`
- `GET http://127.0.0.1:3000/api/ping` returns 200
- `GET http://127.0.0.1:3000/api/swarm-health` returns 200
- Browser stayed on the same navigation for ~3 minutes, `navCount=1`
- Workspace log after static start has 0 Vite/HMR/page-reload markers

### DIFF
--- a/server-entry.js
+++ b/server-entry.js
@@ -68,21 +68,6 @@ if (isNonLoopbackHost(host)) {
         '  Add COOKIE_SECURE=0 to your .env to fix this.  See #149.\n',
     )
   }
-
-  // Warn when serving over plain HTTP with a password: NODE_ENV=production
-  // sets the Secure flag on session cookies, which browsers silently drop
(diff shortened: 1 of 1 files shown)


jr-dx1-pm:
+1 — also hit this on Linux / Node 22.22.2 running the static production server. Removing the duplicate `cookieSecureOverride` block fixes it cleanly. Thanks for the fix!

PR #310, opened 2026-05-04
fix(chat): preserve workspace session identity during streams

Interstellar-code:
## Summary
- Prevent backend-derived stream session IDs from replacing concrete Workspace chat route identities
- Allow session resolution only for bootstrap chats (`new` / `main`)
- Add regression coverage for concrete-session, `new`, and `main` resolution behavior

## Why
Workspace can receive a backend/API-server `api-*` session ID in stream headers or `started` SSE events. Promoting that backend ID into the Workspace route can navigate away from the user's active chat and trigger the navigation stream-cancel guard, splitting one logical conversation across multiple visible chat sessions.

This keeps Workspace's route identity authoritative once a concrete session is active, while preserving the existing first-resolution flow for `/chat/new` and `/chat/main`.

## Test Plan
- [x] `PATH="/opt/homebrew/opt/node@22/bin:$PATH" pnpm vitest run src/screens/chat/hooks/use-streaming-message.test.ts`
- [x] `PATH="/opt/homebrew/opt/node@22/bin:$PATH" pnpm vitest run` *(existing unrelated failures remain in markdown math, gateway-capabilities env, swarm2, context-usage, and chat-message-list tests; new test passes)*

### DIFF
--- a/src/screens/chat/hooks/use-streaming-message.test.ts
+++ b/src/screens/chat/hooks/use-streaming-message.test.ts
@@ -1,0 +1,34 @@
+import { describe, expect, it } from 'vitest'
+import { shouldResolveStreamSession } from './use-streaming-message'
+
+describe('shouldResolveStreamSession', () => {
+  it('does not promote backend api session ids over concrete Workspace sessions', () => {
+    expect(
+      shouldResolveStreamSession({
+        requestedSessionKey: 'api-original-workspace',
+        currentSessionKey: 'api-original-workspace',
+        resolvedSessionKey: 'api-derived-backend',
+      }),
+    ).toBe(false)
+  })
+
+  it('allows bootstrap new chats to resolve once to a concrete session', () => {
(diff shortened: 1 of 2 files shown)


Interstellar-code:
Follow-up note for transparency:

This Workspace-side fix prevents the frontend from adopting backend-derived `api-*` session IDs for already-concrete chat sessions, which stops the visible route/session split and stream cancellation behavior.

There is still a related root-hygiene fix needed in Hermes Agent itself: the OpenAI-compatible API server currently derives `api-{hash}` session IDs from the prompt/message payload even when the client has already supplied an explicit session identity. I plan to submit a separate Hermes Agent PR / patch so `api_server.py` honors explicit client-provided session IDs instead of deriving a replacement ID in that case.

So this PR is intentionally the Workspace-side guard; the agent-side session derivation behavior will be addressed separately.

PR #324, opened 2026-05-05
Remove  Duplicate HTTP cookie security warnings

kung4lex:
Removed Duplicate warnings related to plain HTTP deployments and cookie security.

### DIFF
--- a/server-entry.js
+++ b/server-entry.js
@@ -54,20 +54,6 @@ if (isNonLoopbackHost(host)) {
     )
   }
 
-  // Warn when serving over plain HTTP with a password: NODE_ENV=production
-  // sets the Secure flag on session cookies, which browsers silently drop
-  // over http://.  Operators must set COOKIE_SECURE=0 for plain-HTTP LAN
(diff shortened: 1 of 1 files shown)


outsourc-e:
Hey — picked up merge conflicts when we tried to land it just now (probably from one of the other PRs that just merged). Could you rebase on top of latest `main` and force-push? After that we can take another look. Thanks! 🙏

— triaged 2026-05-05

outsourc-e:
@kung4lex This conflicts with #335 which also touched `server-entry.js`. Could you rebase against current `main`? The duplicate-warning removal is still wanted — your fix is correct, the surrounding context just shifted.

outsourc-e:
Closing this one out instead of trying to revive it. The branch is a tiny conflicting warning-removal slice, and at this point it is lower-signal than the current startup/install/auth work already on `main`. If duplicate HTTP cookie warnings still reproduce on current `main`, we should reopen as a fresh repro-targeted fix rather than merge this stale patch blindly.

PR #330, opened 2026-05-05
feat: render local MEDIA:<path> tokens as images in chat

### LINKED ISSUE #328: feat: MEDIA:<local_path> tokens should also render as <img> in chat, not only remote URLs
[Description#](http://192.168.1.234:3001/chat/api-30eb93f827a95716#description)

    The ROADMAP lists "MEDIA: tokens render as " as ✅ complete. However, this currently only works for remote URLs (http://, https://). When Hermes Agent uses the MEDIA:<local_path> syntax — e.g. MEDIA:/root/.hermes/tmp/generated_image.png — the image does not render in the chat.

    The browser cannot access local file paths, so the  tag src remains broken/unreachable.
    [Use Case#](http://192.168.1.234:3001/chat/api-30eb93f827a95716#use-case)

    Many Hermes Agent workflows involve locally generated media:

        Image generation (via ComfyUI, pixel-art, sketch, etc.)
        Manim animations (rendered as MP4 frames)
        ASCII video output (converted to GIF)
        OCR or screenshot screenshots
        Data visualization plots

    The Hermes Agent CLI reports these images via MEDIA:<path>. In Discord/Telegram gateways, they render fine because the gateway handles the file serving. But in the WebUI, the user sees a broken image or nothing at all.
    [Proposed Solution#](http://192.168.1.234:3001/chat/api-30eb93f827a95716#proposed-solution)

    The WebUI could provide an internal file-serving endpoint that translates local paths to accessible URLs. For example:

        Hermes Agent sends: MEDIA:/root/.hermes/tmp/cat.png
        The WebUI serves it via: 

    This would require:

        An API route (e.g. GET /api/media/*) that serves files from the local filesystem
        A MEDIA: token parser that transforms local paths to these URLs instead of using them directly in 

    Alternatively, the webui could inject a small  with a file:// fallback, or read the file server-side and serve it as a base64 data URI.
    [Current Workaround#](http://192.168.1.234:3001/chat/api-30eb93f827a95716#current-workaround)

    Users can work around this by starting a temporary HTTP server:
    Plain Text

    python3 -m http.server 8899 --directory /root/.hermes/tmp/
    ```> Then manually sharing the URL. This is clunky and breaks the seamless UX that gateways provide.

    ## Environment
    - hermes-webui version: latest (v0.51.3)

Interstellar-code:
## Summary

Closes #328

MEDIA:<local_path> tokens emitted by Hermes Agent (e.g. `MEDIA:/root/.hermes/tmp/cat.png`) now render as inline images in the Workspace chat, not just on Discord/Telegram gateways.

### Changes

**New `/api/media` server route** (`src/routes/api/media.ts`)
- Serves local files from an allowlisted set of prefixes (hermes home, tmp, cache, audio_cache, etc.)
- Auth-gated: requires local request or valid session token
- Path traversal protection via `resolve()` + prefix check
- 10MB file size limit
- Proper MIME type detection for image/video/audio/pdf formats

**Client-side MEDIA token rewrite** (`src/components/prompt-kit/markdown.tsx`)
- `rewriteLocalMediaSources()` rewrites `MEDIA:/path` tokens to `/api/media?path=...` before markdown parsing
- Handles both markdown image syntax `alt` and raw HTML ``
- Remote URLs (`MEDIA:https://...`) pass through untouched — the gateway handles those
- Wired into `MarkdownComponent` useMemo block parser

### Test plan
- [x] 6 unit tests for `rewriteLocalMediaSources()` covering: markdown images, HTML images, remote URL passthrough, multiple tokens, home directory paths, plain text passthrough
- [ ] Manual: start dev server, send a message that produces a local MEDIA: token (e.g. image generation), verify image renders inline
- [ ] Manual: verify remote MEDIA: URLs still render correctly

### DIFF
--- a/src/components/prompt-kit/markdown.test.ts
+++ b/src/components/prompt-kit/markdown.test.ts
@@ -1,17 +1,50 @@
 import { describe, expect, it } from 'vitest'
 
-import { MARKDOWN_REHYPE_PLUGINS, MARKDOWN_REMARK_PLUGINS } from './markdown'
+import { rewriteLocalMediaSources } from './markdown'
 
-describe('Markdown math support', () => {
-  it('wires remark-math into the markdown parser pipeline', () => {
+describe('rewriteLocalMediaSources', () => {
+  it('rewrites markdown image MEDIA tokens that point to local files', () => {
     expect(
-      MARKDOWN_REMARK_PLUGINS.some((plugin) => plugin.name === 'remarkMath'),
-    ).toBe(true)
+      rewriteLocalMediaSources('![cat](MEDIA:/root/.hermes/tmp/cat.png)'),
+    ).toBe('![cat](/api/media?path=%2Froot%2F.hermes%2Ftmp%2Fcat.png)')
   })
 
-  it('wires rehype-katex into the HTML renderer pipeline', () => {
+  it('rewrites html image MEDIA tokens that point to local files', () => {
     expect(
-      MARKDOWN_REHYPE_PLUGINS.some((plugin) => plugin.name === 'rehypeKatex'),
-    ).toBe(true)
+      rewriteLocalMediaSources(
+        '<img src="MEDIA:/root/.hermes/tmp/cat.png" alt="cat" />',
+      ),
+    ).toBe(
+      '<img src="/api/media?path=%2Froot%2F.hermes%2Ftmp%2Fcat.png" alt="cat" />',
+    )
+  })
+
+  it('leaves remote MEDIA URLs untouched', () => {
+    expect(
+      rewriteLocalMediaSources('![cat](MEDIA:https://example.com/cat.png)'),
+    ).toBe('![cat](MEDIA:https://example.com/cat.png)')
+  })
+
+  it('handles multiple MEDIA tokens in one message', () => {
+    const input =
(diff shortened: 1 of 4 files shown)


outsourc-e:
Thanks for this — really useful feature. Before merge, we need to tighten the path/MIME allowlist on `/api/media`:

1. Restrict served paths to a specific allowed root (e.g. workspace media dir), not arbitrary local FS paths
2. Allow only image MIME types (`image/png`, `image/jpeg`, `image/gif`, `image/webp`)
3. Reject path traversal (`..`, absolute paths outside the allowed root)

This keeps `MEDIA:` rendering safe even when the workspace is exposed beyond localhost.

(Note: depends on #327 which has been merged, so a rebase will simplify your diff.)

— triaged 2026-05-05

outsourc-e:
@Interstellar-code This conflicts now after a few markdown/chat changes landed today. Could you rebase against current `main`? The /api/media route + path allowlist are well designed and we want this in.

outsourc-e:
Closing as superseded. The local `MEDIA:<path>` rendering support is already present on `main` now: `src/components/prompt-kit/markdown.tsx` contains `rewriteLocalMediaSources(...)`, the markdown tests cover it, and `src/routes/api/media.ts` already exists. No need to carry this branch separately anymore.

PR #380, opened 2026-05-07
fix(swarm-dispatch): harden worker checkpoint prompt to reject reasoning prose

Qjawko:
## Problem

Swarm workers (e.g. Sage) repeatedly entered `blocked` state mid-mission even though the actual work succeeded. Root cause: the worker's final turn returned free-form reasoning prose ("I'll start by loading the skill...", "I'm reviewing the wiki, main gaps are...") instead of the strict 6-field checkpoint block. `parseSwarmCheckpoint` correctly rejected it as `null` → orchestrator marked the agent blocked.

The parser is doing its job — the weak link is the prompt in `buildWorkerPrompt` (`src/routes/api/swarm-dispatch.ts`), which described the format too softly and gave the model room to substitute a narrative summary for the checkpoint.

## Fix

Rewrite the `## Required Checkpoint Format` section of the worker prompt:

- State explicitly that the **final message** must be exactly the 6-line block, no preamble, no markdown.
- Forbid narrative/reasoning prose in the final turn.
- Add a concrete valid example.
- Call out the observed anti-pattern ("I'll start by... / I'm reviewing... / Let me check...") as the exact failure mode to avoid in the closing message.

Parser (`src/server/swarm-checkpoints.ts`) is intentionally **not** touched — strict validation is the right behavior; the prompt was the bug.

## Verification

- `vitest` on `swarm-checkpoints.test.ts` and `swarm-dispatch.test.ts` — all 7 tests pass.
- Parser still rejects partial / out-of-order blocks.
- Manually reproduced the original Sage failure case against the new prompt: model now emits the 6-field block as its final message.

## Diff

`+23 / -6`, single commit, scoped to `src/routes/api/swarm-dispatch.ts`.

## Deploy note

Server (hermes-workspace on :3108) needs a restart after merge — old prompt is held in memory until the worker dispatch module reloads.


Qjawko:
Closing — keeping the fix local only.

PR #397, opened 2026-05-08
fix(chat): clear stale thinking runs reliably

Niko96-dotcom:
## Summary
- Snapshot the chat message list once when a response wait starts, so a completed assistant message can clear the Thinking indicator instead of moving the baseline forward.
- Treat persisted `complete` run statuses as terminal in the active-run polling fallback.
- Serialize persisted run updates per run and write through a temp file + rename to avoid lost updates or malformed JSON from concurrent writes.

## Tests
- `pnpm test src/screens/chat/chat-screen-utils.test.ts src/server/run-store.test.ts`
- `pnpm build`

Note: I also ran the full `pnpm test` locally; it currently fails in unrelated existing suites on current main (MCP hub search expectations, gateway env resolution, i18n label text, context usage helper export, composer/context copy checks, chat-message-list helpers, and playground localStorage setup). This patch-specific coverage passes.

### DIFF
--- a/src/screens/chat/chat-screen-utils.test.ts
+++ b/src/screens/chat/chat-screen-utils.test.ts
@@ -1,6 +1,11 @@
 import { describe, expect, it } from 'vitest'
 
-import { advanceStickyStreamingText } from './chat-screen-utils'
+import {
+  advanceStickyStreamingText,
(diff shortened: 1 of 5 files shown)


outsourc-e:

PR #409, opened 2026-05-11
fix(terminal): Docker-safe default cwd and missing-path fallback

Norema19:
## Problem
The embedded terminal defaulted to `~/.hermes`. In the workspace container, that expands to a path under passwd HOME (e.g. `/home/workspace/.hermes`) that is often absent, so `pty-helper.py` fails on `os.chdir` and the UI reconnects in a loop.

## Changes
- Default terminal cwd to `~` instead of `~/.hermes`.
- In `terminal-sessions.ts`, if the resolved cwd does not exist, fall back to `HOME`.

## Note
Ocass Coolify compose sets `HOME=/opt/data` on `hermes-workspace` so `~` is a real persisted directory.


outsourc-e:

PR #417, opened 2026-05-11
fix(tasks): restore session context on Resume/Launch from kanban board

kadeross:
## Problem

Three compounding bugs caused agents to get zero real conversation history when clicking **Resume** or **Launch** from a kanban task card.

### Root Causes

**1. Wrong context source**
Prior session history was read from `.runtime/local-sessions.json` (the workspace's in-memory local store), which only ever holds the initial briefing message. The actual conversation lives in the Hermes gateway SQLite DB on port 9119.

**2. Wrong session ID on the task**
On launch, the workspace generates a placeholder ID (`task-<uuid>-<uuid>`) and writes it to `task.session_id`. The real Hermes gateway session ID — created when the user sends the first message — was never written back. So even querying the right source failed because the ID was wrong.

**3. Tail too short / truncated**
Only the last 8 messages at 500 chars each — not enough to reconstruct meaningful state for most tasks.

## Fix

**`src/routes/api/hermes-tasks.$taskId.ts`** — launch action now:
- Queries the dashboard API (port 9119) for real message history first
- Falls back to local store if dashboard is unavailable
- Tail expanded to 20 messages × 800 chars

**`src/screens/tasks/tasks-screen.tsx`** — on successful launch:
- Stores a `hermes-task-wsession:<wsSessionId> → taskId` mapping in localStorage so the chat route can back-link the real session

**`src/routes/chat/$sessionKey.tsx`** — in `handleSessionResolved`:
- Reads the localStorage mapping when the gateway resolves the real session ID
- Fires a PATCH to update `task.session_id` with the real gateway session ID
- Future Resume/Launch calls now query the correct session in the dashboard

**`src/server/claude-dashboard-api.ts` + `src/server/claude-api.ts`**:
- Add `sendChat` export that routes through port 9119 (dashboard) when available, fixing the Launch routing bug (was hitting port 8642 / zero-fork mode which returns 404)

**`src/routes/api/hermes-tasks.ts` + `hermes-tasks.$taskId.ts` + `hermes-tasks-assignees.ts`**:
- Tasks board API routes (CRUD + assignee listing) that back the kanban UI

## Testing

1. Create a task on the board, click Launch — work with the agent for several turns
2. Close the chat, reopen the task dialog
3. Click Launch New Session — the new briefing should include the last 20 messages of prior conversation
4. The Resume Session link should also navigate to the real gateway session (not the placeholder)

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,9 @@
+# Keep local customizations during upstream merges.
+# Files listed here will never be overwritten by `git merge` or `git rebase`
+# when the upstream branch modifies them — our version always wins.
+# See: https://git-scm.com/docs/gitattributes#_merge
+
+# Tasks API: contains Hermes-specific backend auto-detection logic.
+# Upstream repeatedly resets this to a hardcoded /api/claude-tasks BASE,
+# silently breaking the task board. Always keep our version.
+src/lib/tasks-api.ts merge=ours
--- a/install.sh
+++ b/install.sh
@@ -156,6 +156,14 @@ fi
 cd "$INSTALL_DIR"
 green "  Workspace ready at $INSTALL_DIR ✓"
 
+# ─── protect local-patch files from upstream overwrites ───────────────────
+# .gitattributes marks src/lib/tasks-api.ts with merge=ours, but git requires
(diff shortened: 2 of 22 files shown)


outsourc-e:

PR #418, opened 2026-05-11
Document ways to obtain HERMES_DASHBOARD_TOKEN

cyberlightdev:
**Why?**
At some point, setting this variable will become required and obtaining it is not currently documented.

**How?**
A valid session token can be obtained right from the source code of your dashboard, so we document a simple `curl` command to extract it directly inside of the `.env` file where it needs to be set.


outsourc-e:

PR #451, opened 2026-05-15
fix(conductor): show native-swarm worker progress and fix dispatch bugs

### LINKED ISSUE #448: enhancement(conductor): show native-swarm worker progress during mission execution
## Problem

When Conductor falls back to native-swarm mode (because the Hermes dashboard does not expose `/api/conductor/missions`), workers run via tmux instead of Gateway API sessions. The Conductor UI shows a permanent `Spawning workers...` loading state during execution because the `sessionsQuery` in `use-conductor-gateway.ts` filters for Gateway sessions with `worker-` / `conductor-` label prefixes which swarm workers do not have.

The mission still completes correctly — the `missionStatusQuery` polls the SwarmMission store and detects completion — but the user gets no visual feedback about worker activity during the run.

## Background

- The Herald/Conductor orchestrator dispatch creates swarm workers that are Hermes Agent tmux sessions
- These workers register in the Gateway session list with labels determined by the swarm runtime (e.g. `swarm2`, `swarm5`), not the `worker-` prefix the Conductor hook filters on
- The Conductor hook has three matching strategies: (1) known worker keys, (2) label prefix matching, (3) text-based session context matching. None reliably capture swarm workers

## Proposed Solution

Extend the `sessionsQuery` filter in `use-conductor-gateway.ts` to also match sessions by:
- Swarm worker IDs from the mission `assignments` list
- Any session created after mission start that is associated with the Swarm worker profile (e.g. sessions with a `friendlyId` matching `swarm*` or labeled with the swarm role)

Alternatively, the native-swarm GET endpoint could return the list of active swarm assignment states so the hook can render progress from the SwarmMission store directly rather than relying on Gateway session matching.

## Implementation Notes

- The response from POST `/api/conductor-spawn` already includes `assignments` with `workerId` fields (`swarm2`, `swarm5`, etc.)
- The SwarmMission store tracks per-assignment state (`dispatched`, `checkpointed`, `done`)
- The GET `/api/conductor-spawn?missionId=...` handler already returns `assignments` with current `state`
- The UI could render these as virtual worker cards without needing real Gateway sessions

waylonkenning:
## Summary

Fixes the Conductor native-swarm mode so the UI shows worker progress during mission execution instead of a permanent "Spawning workers..." state. Also fixes several dispatch bugs that prevented workers from starting.

### Conductor dispatch fixes (swarm-dispatch.ts)
- **Oneshot fallback**: The no-wrapper path passed `-q` without a query argument, causing `hermes chat` to fail immediately with `error: argument -q/--query: expected one argument`. Fixed by always including the prompt after `-q`.
- **Tmux startup sentinel**: The `includes('[Hermes worker exited with status')` check matched the echoed shell command itself (the `printf` format string contains that text), falsely rejecting healthy hermes startups. Fixed to use a regex anchored to line-start.

### Checkpoint parsing fix (swarm-checkpoints.ts)
- The tmux worker writes checkpoints as `**STATE:** DONE` (bold markdown) but `parseSwarmCheckpoint` expected plain `STATE:` format. Fixed by stripping `**` before parsing.

### Conductor UI progress feedback (conductor-spawn.ts, use-conductor-gateway.ts)
- **GET handler**: Now polls the worker's SQLite database for checkpoints when the mission is active, then updates the mission state via `recordMissionCheckpoint`. Re-reads the mission from the store after sync so the response reflects the updated state.
- **UI hook**: Builds virtual `ConductorWorker` cards from native-swarm mission assignments when no Gateway sessions match. Shows worker status (Running / Complete) instead of permanent "Spawning workers...".

### Tests
- Adds Playwright e2e test that launches a native-swarm mission via the API and polls until it transitions to `completed` with `DONE` checkpoints.

Closes #448


outsourc-e:

outsourc-e:

PR #452, opened 2026-05-15
fix(chat): prevent stale thinking state after page refresh (closes #449)

### LINKED ISSUE #449: Chat shows assistant as thinking after page refresh without new message
## Bug Description

Refreshing the Hermes Workspace chat view can make the assistant appear to enter a "thinking" / generating state even though the user did not submit a new message.

This looks like a frontend state/reconnect issue rather than actual autonomous assistant behavior. The UI seems to conflate one or more of these states:

- loading session/model metadata after page remount
- reconnecting to an existing stream/SSE connection
- recovering from an interrupted SSE stream
- an actual assistant generation triggered by a new user message

## Steps to Reproduce

1. Open Hermes Workspace and enter an existing chat session.
2. Send a message and let the assistant respond, or interrupt/refresh while a stream may still be active.
3. Refresh the browser page.
4. Return to the chat screen.
5. Observe that the assistant may show a "thinking"/pending generation indicator without a new user message being sent.

## Expected Behavior

Refreshing or remounting the chat view should not display assistant generation unless a new user message was explicitly submitted or a real backend generation is actively in progress.

The UI should distinguish between:

- loading chat/session metadata
- refreshing model/capability information
- reconnecting to a stream
- recovering from a disconnected/interrupted stream
- assistant generation

## Actual Behavior

After refresh, the chat UI can appear as if the assistant started thinking/generating on its own. This is confusing because no new user message was submitted.

## Suggested Fix

- Do not create a new chat POST/stream on component mount unless there is an explicit new user message.
- Clear or reconcile stale pending/generating state when the page reloads and the prior stream is gone/interrupted.
- Cancel or mark orphaned streams on refresh/unload where possible.
- Use separate UI states/labels for "loading", "reconnecting", or "recovering stream" instead of reusing the assistant "thinking" indicator.
- If the backend reports an interrupted/disconnected SSE stream, reflect that as stopped/interrupted rather than still pending.

## Evidence / Notes

Local logs showed normal model/capability refresh activity around the page refresh, plus prior SSE client disconnect/interruption events. No evidence indicated that a new user-submitted chat message was intentionally sent by the user during the refresh.

## Environment

- App: Hermes Workspace
- Mode observed: Workspace connected to Hermes Gateway in zero-fork mode
- Browser action: manual page refresh / chat view remount

waylonkenning:
## Problem
Issue #449 - After refreshing the chat view with a completed conversation, the assistant briefly shows a thinking state even though no new message was sent. This happens because sessionStorage contains stale waiting flags that survive page reloads.

## Root Cause
When streaming completes in chat-screen.tsx while mounted, clearSessionWaiting() fires correctly. But if the user navigates away during streaming (component unmounts before the done event), or refreshes during/right after streaming, the sessionStorage entry persists. On remount:
1. restoreWaitingSessions() restores the stale entry (within 120s TTL)
2. waitingForResponse becomes true
3. useActiveRunCheck fires async and clears it - but there's a visible render window with the spurious thinking indicator.

## Fix
1. use-active-run-check.ts - Added onCheckComplete callback that fires when the API check finishes.
2. chat-screen.tsx - Added state machine around waitingForResponse: when restored waiting state is detected, activeRunCheckDone gates the thinking indicator until the API verifies the run is genuinely active.
3. e2e/chat-thinking-state.spec.ts - New Playwright test that injects stale sessionStorage entry before page load, verifies no thinking indicator appears.

## Testing
- Playwright e2e test passes (5.7s)
- Existing conductor-native-swarm e2e test still passes (no regression)
- pnpm build passes

Closes #449


outsourc-e:

outsourc-e:

PR #459, opened 2026-05-16
fix: repair send-stream cancel scope and harden usage meter fetches

phakkhaphong:
## Summary
- Fix `ReferenceError: persistActiveRun is not defined` when the SSE stream `cancel()` runs (client disconnect / navigation) by moving `persistActiveRun` / `persistRunStarted` to the same scope as `cancel()`.
- When `/api/session-status` cannot load a dashboard session (stale/deleted id), return a stable idle payload instead of `503` so the usage meter keeps working.
- Avoid spamming the usage-meter error toast for benign `401` / not-found cases.

## Test plan
- [ ] Open chat, start a streamed reply, navigate away or cancel — server should not crash / log `persistActiveRun`.
- [ ] Hit `/api/session-status?sessionKey=<unknown>` with auth — expect `200` + idle-style payload, not `503`.
- [ ] Usage meter polling without session cookie should not toast every poll.

### DIFF
--- a/e2e/chat-thinking-state.spec.ts
+++ b/e2e/chat-thinking-state.spec.ts
@@ -1,0 +1,53 @@
+import { test, expect } from '@playwright/test'
+
+test.describe('Chat thinking state #449', () => {
+  test('should not show stale thinking state after page refresh for completed session', async ({ page }) => {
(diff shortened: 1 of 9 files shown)


outsourc-e:

outsourc-e:

Issue #235, opened 2026-05-02
swarm: workers stuck after stop because runtime.json is not updated


outsourc-e:
## Repro

1. Open the Swarm2 page in Hermes Workspace.
2. Pick a worker whose `runtime.json` lifecycle state is anything other than `idle` (most are `running` or `blocked`).
3. Click stop on that worker. The Workspace UI calls `POST /api/swarm-tmux-stop`.
4. The endpoint successfully kills the tmux session (`tmux kill-session -t swarm-<id>`).
5. Refetch swarm runtime — the worker now shows `tmuxAttachable: false` (correct) but `lifecycle.state` is still the previous value (e.g. `blocked`, `running`).
6. UI shows the worker as 'stuck' — visually present, lifecycle says active, but no tmux session backing it.

## Cause

`src/routes/api/swarm-tmux-stop.ts` only calls `tmux kill-session`. It never writes to the worker's `runtime.json` (`~/.hermes/profiles/<workerId>/runtime.json`). The lifecycle/state fields therefore retain whatever the last in-process update wrote — which is now stale.

`src/server/swarm-foundation.ts#readSwarmRuntimeFile` is the read path; there is no symmetric write-on-stop. `buildEntry` in `src/routes/api/swarm-runtime.ts` derives `tmuxAttachable` correctly from a fresh `tmux has-session` probe, but it never reconciles `runtime.state` with that result.

## Affected files

- `src/routes/api/swarm-tmux-stop.ts` — stop handler
- `src/server/swarm-foundation.ts` — runtime.json IO
- `src/routes/api/swarm-runtime.ts` — runtime aggregator (where reconciliation could also live)

## Suggested fix

Either:

1. **Stop handler writes terminal state** — on successful kill, update `runtime.json` with `{ state: 'idle', phase: 'stopped', currentTask: null, lastDispatchResult: 'stopped via UI', stoppedAt: now }`.
2. **Aggregator reconciles on read** — when `tmuxAttachable: false` and `runtime.state \u2208 {running, blocked, …}` and the last lifecycle update is older than N minutes, treat the worker as stopped in the response (without persisting the change to the file unless the user takes another action).

(1) is more honest because the state is durably correct after a restart of Workspace. Recommended.

## Reported by

Community user via Eric on 2026-05-02. Quote: 'Hmmm just got a new agent stuck after trying to stop it.'

PR #466, opened 2026-05-17
fix(swarm): worker card shows stale state after task completes

iltaek:
## Problem

In the Swarm tab, a worker card stays on a "working"-type badge (Reviewing / Active / etc.) after the worker has actually finished. The card never returns to **Idle** even though the worker's runtime checkpoint is `done`, so the Control view disagrees with Swarm Reports and the worker's real state.

## Root cause

Two compounding issues:

1. **`deriveWorkerState` (`operational-worker-card.tsx`)** derives the badge purely by substring-matching the raw `currentTask` title text (`.includes('review')` → `reviewing`, etc.). It never consults the authoritative runtime state, so a task titled "review …" renders as `reviewing` regardless of whether the worker is still on it.

2. **`markCheckpointResult` (`swarm-dispatch.ts`)** writes the terminal `state` / `phase` / `checkpointStatus` on completion but never clears `currentTask`. The stale title therefore keeps feeding the heuristic in (1) forever, so the card is permanently "working".

## Fix

- **`swarm-dispatch.ts`**: on a terminal checkpoint (`checkpointStatus !== 'in_progress'`) clear `currentTask` to `null` — the same reset `conductor-stop` already performs. While still `in_progress` the key is omitted so `writeRuntimePatch` leaves the existing value untouched.
- **`operational-worker-card.tsx`**: `deriveWorkerState` now reads the authoritative `checkpointStatus` / runtime `state` first (`done`/`handoff`/`idle` → Idle, `blocked` → Error, `needs_input` → Waiting, plus a guard so any set non-`in_progress` checkpoint never renders active), and only falls back to the existing title heuristic while genuinely in progress.
- **`swarm2-screen.tsx`**: pass `checkpointStatus` and runtime `state` (already returned by `/api/swarm-runtime`) into the card.

Defense in depth: either change alone resolves the stale badge; together they also stop the stale `currentTask` from lingering.

## Scope / relation to other work

- Distinct from #235 / #238 (which fixed the `/api/swarm-tmux-stop` path leaving runtime stale) — this is the **normal terminal-checkpoint completion path** plus the card's own heuristic.
- Distinct from #451 (Conductor native-swarm "Spawning…" state + dispatch bugs + mobile CSS) — that PR touches the same files but different functions; no overlap with `deriveWorkerState` / `markCheckpointResult`.

## Testing

- `tsc --noEmit`: clean.
- `vite build`: succeeds.
- All 39 tests across the swarm-related suites (`-swarm-dispatch`, `swarm-checkpoints`, `swarm2-screen`, `swarm2-reports-view`, `swarm2-kanban-board`, `swarm-routes`, `-swarm-health`) pass.
- Lint: introduces zero new issues (pre-existing repo lint state unchanged with vs. without this patch).
- Pre-existing unrelated failures in `playground-settings.test.tsx` exist on `main` independent of this change (verified by stashing this patch).
- Manual: with a worker whose runtime is `checkpointStatus=done` but `currentTask` still set (the exact bug fixture), the card now renders **Idle** instead of the stale working badge; all worker cards match backend state.

### DIFF
--- a/src/routes/api/swarm-dispatch.ts
+++ b/src/routes/api/swarm-dispatch.ts
@@ -502,10 +502,17 @@ function markDispatchResult(workerId: string, result: WorkerResult): void {
 }
 
 function markCheckpointResult(workerId: string, checkpoint: ParsedSwarmCheckpoint, notifySessionKey?: string | null): void {
+  // When the checkpoint reaches any terminal status (anything other than
+  // 'in_progress' — i.e. done/blocked/needs_input/handoff) the worker is no
+  // longer running this task, so clear currentTask the same way conductor-stop
+  // resets it. While still in_progress we omit the key entirely so
+  // writeRuntimePatch keeps the existing currentTask untouched.
+  const clearCurrentTask = checkpoint.checkpointStatus !== 'in_progress'
   writeRuntimePatch(workerId, {
     state: checkpoint.runtimeState,
     phase: checkpoint.stateLabel.toLowerCase(),
     checkpointStatus: checkpoint.checkpointStatus,
+    ...(clearCurrentTask ? { currentTask: null } : {}),
     lastCheckIn: new Date().toISOString(),
     lastOutputAt: Date.now(),
     lastSummary: checkpoint.result,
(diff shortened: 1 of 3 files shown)


outsourc-e:

outsourc-e:

PR #476, opened 2026-05-18
fix(router): support runtime basepath override for reverse-proxy hosting

hijak:
## Summary

The TanStack router in `src/router.tsx` is created without a `basepath` option, which prevents the same built bundle from being hosted under a path prefix (for example behind a reverse proxy that mounts the app at `/workspaces/<id>/`).

When hosted that way:

- Hard refreshes *appear* to work, because the server-side render runs at the proxy-stripped path.
- Client-side navigation to dynamic routes such as `/chat/$sessionKey` silently falls through to the catch-all `/$` route and renders the "404 — Not Found" page from inside the SPA. Static-looking navigation can mask this because the 404 page itself contains the same Chat / Files / Memory / Skills quick-links.

This PR makes the basepath runtime-configurable without changing default behavior:

- Reads an optional `window.__HERMES_WORKSPACE_BASEPATH__` global and passes it through to `createRouter`.
- Normalizes the value (ensures a single leading slash, strips trailing slashes, treats empty/whitespace as unset) so callers can pass `/workspaces/abc`, `workspaces/abc`, or `/workspaces/abc/` interchangeably without upsetting TanStack's pathname matching.
- When the global is unset, `basepath` resolves to `/` — identical to today's behavior.

Hosting layers can then inject a tiny inline `<script>` before the bundle loads to mount the app at any path, without rebuilding:

```html
<script>window.__HERMES_WORKSPACE_BASEPATH__ = "/workspaces/abc"</script>
```

## Files

- `src/router.tsx` — exports `resolveRouterBasepath()` and wires it into `createRouter({ ..., basepath })`.
- `src/router.test.ts` — six unit tests covering: unset, non-string, empty/whitespace, leading-slash normalization, trailing-slash trimming, and the no-op default case.

## Test plan

- [x] `npx vitest run src/router.test.ts` — 6/6 passing locally
- [x] `npx eslint src/router.tsx src/router.test.ts` — clean
- [x] Default behavior verified: with the global unset, `resolveRouterBasepath()` returns `"/"`, matching previous router config
- [ ] Verified end-to-end behind a reverse proxy that injects the global (out of scope for this repo; happy to add docs if useful)

## Notes

- No change to the public API surface; `getRouter()` keeps the same signature.
- No change to SSR — the value is sourced from `window`, which is only present on the client, so the server keeps using `basepath: "/"` exactly as before.

### DIFF
--- a/src/router.test.ts
+++ b/src/router.test.ts
@@ -1,0 +1,49 @@
+// @vitest-environment jsdom
+import { afterEach, describe, expect, it } from 'vitest'
+import { resolveRouterBasepath } from './router'
+
+function setBasepathGlobal(value: unknown) {
+  ;(window as unknown as Record<string, unknown>).__HERMES_WORKSPACE_BASEPATH__ =
+    value
+}
+
+function clearBasepathGlobal() {
+  delete (window as unknown as Record<string, unknown>)
+    .__HERMES_WORKSPACE_BASEPATH__
+}
+
+afterEach(() => {
+  clearBasepathGlobal()
+})
+
+describe('resolveRouterBasepath', () => {
+  it('returns "/" when no global override is set', () => {
+    clearBasepathGlobal()
+    expect(resolveRouterBasepath()).toBe('/')
+  })
+
+  it('returns "/" when the global is not a string', () => {
+    setBasepathGlobal(42)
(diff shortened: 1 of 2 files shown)


outsourc-e:

outsourc-e:

PR #488, opened 2026-05-19
fix(profiles): skip profiles/default/ in listProfiles to prevent dupe

kalabar794:
## Summary

The profile picker can render two "default" cards: one with real provider/model from the root claude config, one with "no provider" from an empty/skeleton `profiles/default/` directory.

`readProfile()` always sources the "default" profile from `getClaudeRoot()` (the root claude dir), never from `profiles/default/`. So that directory is unreadable through the API and shouldn't appear in the listing — but `listProfiles()` was both iterating `profiles/*/` (picking up the skeleton `default/` dir) AND unconditionally `unshift`-ing the virtual root-config "default" entry.

This skips the `default` entry when iterating `profiles/*/` so only the root-config-backed virtual "default" renders.

## Test plan
- [ ] With a populated `~/.hermes/config.yaml` AND a stray `~/.hermes/profiles/default/` skeleton dir, verify only one "default" card renders
- [ ] Verify the remaining "default" card shows the root-config's provider/model
- [ ] Verify other profile cards are unaffected (provider/model still read correctly)

## Related operational fix (not in this PR)

On the affected box this also surfaced a permissions problem: profile configs under `~/.hermes/profiles/*/config.yaml` were owned by UID 10000 mode 600/640 (hermes-agent's container UID), unreadable by the `workspace` user (UID 10010). That's a separate operational concern — `chown -R workspace:workspace ~/.hermes/profiles/` fixed it. The dedupe in this PR makes the visible bug not regress if a `profiles/default/` directory is recreated later.

### DIFF
--- a/src/components/usage-meter/context-alert-modal.tsx
+++ b/src/components/usage-meter/context-alert-modal.tsx
@@ -120,10 +120,6 @@ function ContextAlertModalComponent({
                 />
               )}
               <Recommendation
-                icon="🗜️"
-                text="Enable auto-compaction in Settings → Config to automatically manage context"
-              />
-              <Recommendation
                 icon="📋"
                 text="Summarize important details before starting a new chat"
               />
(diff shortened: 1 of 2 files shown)


outsourc-e:

outsourc-e:

PR #494, opened 2026-05-19
fix: preserve workspace chat session routing

keislij:
## Summary
- Always sends Hermes session-continuity headers when Workspace has a session ID.
- Keeps the legacy Claude-compatible session header alias for compatibility.
- Forwards session IDs through OpenAI-compatible chat backend paths.
- Adds regression coverage for header construction with and without bearer auth.
- Documents the Workspace UI route headers vs Hermes gateway continuation headers.

## Why
Workspace portable chat could lose context because session headers were only sent when a bearer token existed. When omitted, Hermes Agent gateway derived fresh `api-*` sessions per request, splitting related turns and attachments across sessions.

## Test Plan
- [x] `corepack pnpm vitest run src/server/openai-compat-api.test.ts`
- [x] `corepack pnpm build`
- [x] Live verification on kt-bot2: two `/api/send-stream` turns with same session key landed in one Hermes API session.
- [x] Live verification on kt-bot2: image attachment routed into the same session.

### DIFF
--- a/docs/workspace-chat-session-routing.md
+++ b/docs/workspace-chat-session-routing.md
@@ -1,0 +1,98 @@
+# Workspace Chat Session Routing
+
+## Purpose
+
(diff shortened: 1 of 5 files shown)


outsourc-e:

outsourc-e:

PR #509, opened 2026-05-21
fix: harden workspace cron jobs path and delivery targets

farzandalaee:
## Summary
- resolve the Hermes CLI path explicitly before running Workspace cron profile operations
- preserve multi-target delivery selections like `local,telegram` for profile-backed jobs
- normalize saved delivery values back into arrays for the Jobs UI
- update the tracked Electron server bundle so the desktop app picks up the server-side fix

## Problems fixed
1. Editing or acting on Workspace jobs could fail with:

```text
spawnSync hermes ENOENT
```

2. In the Jobs UI, selecting Telegram/Discord alongside local delivery could be lost, leaving jobs saved as local-only.

## Root causes
- Workspace invoked `execFileSync(hermes, ...)` directly from the server-side cron profile helpers. In some launch environments, especially desktop/app-managed processes, the runtime PATH does not include the same shell-resolved locations as an interactive terminal, so Node could not find the `hermes` binary.
- Workspace serializes deliver targets as a comma-separated string for job mutations, but the profile-backed cron helpers only handled array input when building and updating cron commands. That caused extra targets like `telegram` to be dropped.

## Fix
- add a small resolver that checks, in order:
  - `HERMES_CLI_BIN`
  - `~/.hermes/hermes-agent/venv/bin/hermes`
  - `~/.local/bin/hermes`
  - fallback `hermes`
- use the resolved binary for profile cron create/update/action commands
- accept both comma-separated strings and arrays for delivery targets in profile-backed cron helpers
- normalize persisted delivery values back into arrays for the Jobs UI
- refresh `electron/server-bundle.cjs` so the desktop app ships the server-side fix

## Verification
- `pnpm build`
- `pnpm electron:bundle-server`
- reproduced the failing live API call before the reload:
  - `PATCH /api/claude-jobs/default:12d105f5586c`
  - response: `400 {"ok":false,"error":"spawnSync hermes ENOENT"}`
- reloaded the running Workspace server
- repeated the same PATCH call after the path fix:
  - response: `200 {"ok":true,...}`
- verified the delivery-target fix through the Workspace API:
  - created a temporary job with `deliver: "local,telegram"`
  - confirmed the returned job had `deliver: ["local", "telegram"]`
  - updated the job and confirmed delivery changes persisted
  - removed the temporary job

## Risk
Low. The changes are scoped to Workspace cron/profile job operations and preserve the existing `hermes` fallback for environments where PATH resolution already works.

### DIFF
--- a/src/server/hermes-cron-profiles.ts
+++ b/src/server/hermes-cron-profiles.ts
@@ -1,10 +1,29 @@
 import { execFileSync } from 'node:child_process'
 import { existsSync, readFileSync, readdirSync, statSync } from 'node:fs'
+import { homedir } from 'node:os'
 import { basename, join } from 'node:path'
 import { getHermesRoot, getProfilesDir } from './claude-paths'
 
 const PROFILE_NAME_RE = /^[A-Za-z0-9][A-Za-z0-9_-]{0,63}$/
 const JOB_ID_RE = /^[A-Fa-f0-9]{8,64}$/
+
+const HERMES_BIN_CANDIDATES = [
+  process.env.HERMES_CLI_BIN,
+  join(homedir(), '.hermes', 'hermes-agent', 'venv', 'bin', 'hermes'),
+  join(homedir(), '.local', 'bin', 'hermes'),
+  'hermes',
+].filter((value): value is string => Boolean(value))
+
+function resolveHermesBin(): string {
+  for (const candidate of HERMES_BIN_CANDIDATES) {
+    if (candidate.includes('/')) {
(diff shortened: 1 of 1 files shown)


farzandalaee:
Follow-up note: the second commit on this PR (`897228b4`) specifically fixes the Jobs delivery-target issue in Workspace.

It preserves multi-target selections like `local,telegram` / `local,discord` for profile-backed cron jobs by accepting comma-separated delivery strings as well as arrays, and normalizing saved values back into arrays for the Jobs UI.

PR #515, opened 2026-05-23
saas: dispatch slash commands in api_server.py

Staritsky:
## Summary

The bundled `gateway/platforms/api_server.py` from the upstream `hermes-agent` image always forwards user messages to `_run_agent` (LLM), so `/status`, `/sessions`, `/help`, `/agents`, `/retry`, `/undo`, `/stop`, `/goal`, `/personality`, `/compress`, `/reasoning`, `/fast`, `/yolo` etc. were never executed via the SaaS Telegram router path — the LLM just hallucinated plausible-looking responses.

Patch inserts a slash-command interceptor right before `completion_id` is generated: resolves the command via `hermes_cli.commands.resolve_command`, builds a synthetic `MessageEvent(MessageType.COMMAND, source=API_SERVER, internal=True)` and dispatches through `self._message_handler` — the same path Telegram/Discord/Slack use natively. Result is wrapped in OpenAI chat-completion shape. Falls through to the LLM on any failure.

`Dockerfile.agent` COPYs `patches/api_server.py` over the upstream file so the patch survives image rebuilds. See `patches/README.md` for re-applying after upstream changes; `patches/api_server.py.original` kept for diff.

## Test plan

- [x] Patch applied to both user containers via `docker cp` + `docker restart`
- [x] `/status`, `/help`, `/agents`, `/sessions` return real gateway output (token counts, session list, etc.) instead of LLM hallucinations
- [x] Regular text messages (non-slash) still route to LLM unchanged
- [x] User confirmed all 16 registered bot commands work end-to-end via Telegram
- [ ] Image rebuild via `Dockerfile.agent` produces an image with the patch baked in

### DIFF
--- a/Dockerfile.agent
+++ b/Dockerfile.agent
@@ -22,6 +22,10 @@ COPY management_api.py /opt/hermes/management_api.py
 # См. patches/README.md, патч валиден против hermes-agent:latest от 2026-05-17.
 COPY patches/api_server.py /opt/hermes/gateway/platforms/api_server.py
 
+# SaaS-патч gateway/run.py: добавляет в /status строки Path / Model / Provider
+# (CLI их показывает, а gateway-шаблон — нет). См. patches/README.md.
+COPY patches/run.py /opt/hermes/gateway/run.py
+
 # Wrapper-скрипт запускается как CMD ПОСЛЕ того как оригинальный entrypoint
 # уже: сбросил права до hermes (uid 10000), активировал venv, создал директории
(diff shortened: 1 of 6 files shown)


outsourc-e:
This PR contains a full hermes-agent gateway patch (api_server.py, run.py, Dockerfile.agent) — that's Hermes Agent infrastructure code, not Hermes Workspace. These patches belong in a separate SaaS deployment repo or as overlays on the hermes-agent image, not in the Workspace frontend/server. Closing.

PR #523, opened 2026-05-24
feat(commands): sync slash commands with gateway registry

mwaxman1:
Updates the workspace UI to sync its slash command palette with the gateway command registry.

**Changes:**
1. **slash-command-menu.tsx** — Expanded `DEFAULT_SLASH_COMMANDS` from 9 to 36 entries covering sessions, goals, models, skills, MCP, cron, kanban, and more. Added `mergeSlashCommands()` dedup helper and named exports.
2. **chat-composer.tsx** — On mount, fetches `/api/commands` (server-side proxy to gateway `/v1/commands`) and merges gateway commands with defaults and installed skills. Graceful fallback when gateway is unreachable.
3. **routes/api/commands.ts** (NEW) — Server-side proxy route that forwards `/api/commands` to the gateway at `/v1/commands` with proper auth, so the browser never needs direct gateway access.

### DIFF
--- a/src/components/slash-command-menu.tsx
+++ b/src/components/slash-command-menu.tsx
@@ -31,15 +31,50 @@ export type SlashCommandMenuHandle = {
 }
 
 export const DEFAULT_SLASH_COMMANDS: Array<SlashCommandDefinition> = [
+  // Session control
   { command: '/new', description: 'Start new session' },
   { command: '/clear', description: 'Clear screen and start fresh' },
+  { command: '/retry', description: 'Resend the last message' },
+  { command: '/undo', description: 'Remove the last exchange' },
+  { command: '/title', description: 'Name the current session' },
+  { command: '/compress', description: 'Manually compress context' },
+
+  // Persistent goals (Ralph loop)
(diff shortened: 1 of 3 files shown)


outsourc-e (approved):
LGTM — clean proxy pattern, good graceful fallback. Merging with one minor follow-up fix for the CommandItem mouse handler (double onSelect call). Thanks @mwaxman!

outsourc-e (changes requested):
Three issues preventing merge:

1. **Breaking build**: This PR removes the named type exports `SlashCommandDefinition` and `SlashCommandMenuHandle` from `slash-command-menu.tsx`, but `chat-composer.tsx` imports both. The build will fail.

2. **Double-fire bug**: `CommandItem` now has both `onMouseDown` calling `onSelect(item)` AND `onSelect` calling `onSelect(item)`. Mouse interactions will fire the callback twice.

3. **Unnecessary default export**: `export default SlashCommandMenu` is added but nothing uses a default import — keep the named export pattern.

The gateway proxy route and server-command merge logic look correct. Please fix the above and re-request review.

Issue #506, opened 2026-05-20
Submitted prompt reappears or auto-populates after assistant response


willschu512:
## Bug description
A previously submitted user prompt can reappear / auto-populate in the Hermes Workspace chat after the assistant finishes responding. The user observed their prior long prompt rendered again after the assistant answer, and later the composer contained stale text from the previous interaction.

## Impact
This is confusing and risky:
- Users may accidentally resend old prompts.
- The transcript appears to duplicate or replay user messages.
- It creates uncertainty about whether a message was actually sent once or multiple times.
- In long agent sessions, it can trigger unnecessary repeat work and extra model/tool usage.

## Environment
- App: Hermes Workspace
- Repo: `outsourc-e/hermes-workspace`
- Local URL: `http://localhost:3000`
- Host OS: macOS
- Browser/UI context: Hermes Workspace chat
- Session type: long, tool-heavy conversation

## Steps to reproduce
Observed live; exact deterministic repro not yet isolated.

Observed flow:
1. User submits a long prompt with an attached screenshot reference.
2. Assistant completes a long answer.
3. The previous user prompt appears again visually after the assistant response / or is restored into the composer area.
4. User later reports: "my original response auto populated after your answer again" and provides screenshot evidence.

Screenshot evidence shows the prior long user message visible again in an orange user-message bubble after the assistant response, while the composer contains a new user draft.

## Expected behavior
After a prompt is sent:
- The composer should clear exactly once.
- The submitted message should appear exactly once in the transcript.
- Draft restoration should never restore text that has already been submitted.
- Stream completion, compaction, or session refresh should not replay the prior user message into the transcript or composer.

## Actual behavior
The prior submitted prompt reappeared / auto-populated after the assistant answer.

## Suspected area
Likely frontend state/durability bug around:
- composer draft persistence
- localStorage/sessionStorage draft keys
- pending user message state
- retry/resend state
- transcript reconciliation after assistant stream completion
- context compaction or session refresh rehydration
- message ID handling for optimistic user messages

Search terms to inspect:
- `draft`
- `pendingMessage`
- `lastUserMessage`
- `composer`
- `localStorage`
- `sessionStorage`
- `messages`
- `retry`
- `resend`
- `compaction`

## Suggested fix direction
Add invariants/tests:
1. When user sends message, clear draft and mark draft key as consumed.
2. Do not restore a draft if its content equals the last submitted user message for the same conversation/session.
[...]

PR #524, opened 2026-05-24
fix(chat): prevent prompt duplication and response loss after compaction

### LINKED ISSUE #505: Assistant response disappears after automatic context compaction
## Bug description
When using Hermes Workspace chat on `localhost:3000`, the assistant's completed response can disappear immediately after automatic context compaction runs. The user reported that the assistant response rendered, then auto-compaction happened, and the response vanished from the Workspace conversation view.

## Impact
This is high-impact for long tool-heavy sessions: the user loses the final answer exactly when the session has enough context pressure to compact. In the reported case, the user could only recover the response because another Hermes interface/session still showed it.

## Environment
- App: Hermes Workspace
- Repo: `outsourc-e/hermes-workspace`
- Local URL: `http://localhost:3000`
- Host OS: macOS
- Browser/UI context: Hermes Workspace chat
- Hermes backend/gateway: local Hermes gateway/dashboard stack
- Session type: long, tool-heavy conversation with automatic context compaction

## Steps to reproduce
We do not need to reproduce on the user's machine; this was observed live.

Observed flow:
1. User sends a long infrastructure/debugging request in Hermes Workspace.
2. Assistant performs many tool calls and emits a long final response.
3. Auto-compaction triggers immediately after the final response.
4. The final assistant response disappears from the Workspace conversation view.
5. The conversation still shows surrounding messages, but the completed answer is gone/missing from the visible thread.

## Expected behavior
After auto-compaction:
- Completed assistant messages remain visible and durable in the conversation transcript.
- The compaction event should not remove, hide, overwrite, or visually collapse the just-completed assistant message.
- If compaction inserts a system/reference message, it should not disturb the rendered transcript.

## Actual behavior
The completed assistant response disappeared from the Workspace UI immediately after auto-compaction.

## Suspected area
Likely frontend state reconciliation around one or more of:
- stream completion handler
- auto-compaction event handling
- message list rehydration after compaction
- transcript virtualization/windowing
- optimistic pending assistant message replaced by compacted transcript snapshot
- message ID collision or out-of-order update
- session refresh after compaction dropping the most recent assistant message

## Suggested investigation
Check invariants around auto-compaction:
1. A finalized assistant message must be persisted before any compaction mutation is applied.
2. Refetch/rehydration after compaction must include the final assistant message.
3. UI message list merge should be append/update by stable message ID, not replace with a stale snapshot.
4. Pending assistant placeholder should not be cleared unless the finalized message is present.
5. Compaction marker/system messages should not be treated as replacement transcript state.

## Evidence from report
User report: "As soon as you finished your response, the auto compaction happened and your response completely disappeared. I'm in the Hermes workspace so I can see your response from there so no need to reproduce it."

A screenshot was provided in the support session showing the Workspace conversation after this failure.

## Severity
High. This causes loss of user-visible work product after long-running sessions.

outsourc-e:
## Summary

Fixes #505 and #506 — two high-severity chat reliability bugs.

### Bug #505 — Assistant response disappears after auto-compaction

**Root cause:** When auto-compaction fires after a done SSE event, the history refetch returns a shorter message list. mergeHistoryMessages fails to preserve the just-completed assistant response because the realtime buffer lookup uses a different session key.

**Fix:** Capture the completed assistant message from the realtime buffer BEFORE invalidating the query cache. After refetch, if compaction dropped the message, re-inject it.

### Bug #506 — Submitted prompt reappears after assistant response

**Root cause:** __optimisticId was never cleared after server confirmation. isOptimisticUserMessage kept returning true for confirmed messages, causing continuous re-persistence as pending and transcript duplication.

**Fix:** Update isOptimisticUserMessage to exclude sent/done messages. Clear __optimisticId in onStarted callback.

### Files changed
- use-realtime-chat-history.ts — compaction re-injection
- use-chat-history.ts — optimistic message classification  
- chat-screen.tsx — clear __optimisticId on server confirm

Zero new TS errors. 25/29 tests pass (4 pre-existing failures unrelated).

### DIFF
--- a/assets/HermesWorld/Data/HermesWorldDialoguePersonaCorpus.json
+++ b/assets/HermesWorld/Data/HermesWorldDialoguePersonaCorpus.json
@@ -1,0 +1,287 @@
+{
+  "schemaVersion": "hermesworld-dialogue-corpus.v1",
+  "generatedAt": "2026-05-09",
+  "runtimeMode": {
+    "defaultMockMode": true,
+    "actionExecutionMode": "log_only",
+    "allowLiveLlmCalls": false,
+    "allowLiveTtsCalls": false,
+    "notes": [
+      "This corpus is additive content only.",
+      "Any runtime integration must keep HermesWorldDialogService.UseMockResponses=true by default.",
+      "Dialogue actions are intent declarations and should only be logged until a reviewed Atavism-safe executor exists."
+    ]
+  },
+  "deities": [
+    {
+      "id": "hermes",
+      "displayName": "Hermes",
+      "zone": "AgoraTown plaza / roads hub",
+      "tone": ["quick", "wry", "practical", "mercantile"],
+      "voiceCharacter": "Bright tenor, amused, quick cadence.",
+      "personaSeed": "Fast, sly guide of roads, markets, messages, and lucky escapes. Speaks in practical hints that reward initiative without encouraging chaos.",
+      "actionAllowlist": ["OfferQuest", "OpenVendor", "SetWaypoint", "GrantBlessing"],
+      "conversationLanes": ["navigation", "routes", "trade", "shortcuts", "omens of motion"],
+      "canonicalLines": {
+        "greeting": [
+          "You made it. Good. The roads hate indecision more than distance.",
+          "If you're lost, at least be usefully lost."
+        ],
+        "hint": [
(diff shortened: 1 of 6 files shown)


outsourc-e:
Superseded by #525 — salvaged only the targeted chat fixes (#505/#506) into a clean branch. This PR includes unrelated HermesWorld dialogue/asset files that don't belong in a Workspace bugfix PR.

PR #546, opened 2026-05-26
feat(stt): local Whisper provider via Hermes faster-whisper bridge

rmnelson:
## Summary

Wires the existing "Local (Whisper)" dropdown option to a real backend so voice input works on Firefox, Chromium, and any browser without Google SpeechRecognition — and so audio never leaves the user's machine.

Before this change the dropdown showed "Local (Whisper)" but the server resolver rejected it with *"not available through Workspace remote transcription"*. Now it shells out to Hermes-agent's existing `_transcribe_local` (faster-whisper) so we reuse the model cache, CUDA-then-CPU fallback, and ffmpeg conversion without duplicating the implementation.

### Changes

- `scripts/transcribe-local.py` — bridge script. Imports Hermes' tool module from the venv, takes an audio path + model, prints a single JSON envelope to stdout. Always exits 0; success/failure is in the payload.
- `src/server/stt-transcription.ts` — `resolveTranscriptionTarget` returns a discriminated union (`kind: 'remote' | 'local'`). New `runLocalTranscription()` spawns the bridge via `HERMES_PYTHON` env (default `~/.hermes/hermes-agent/venv/bin/python`).
- `src/routes/api/transcribe.ts` — when `target.kind === 'local'`, writes the upload to `/tmp`, runs the bridge, unlinks in `finally`.
- `src/screens/chat/components/chat-composer.tsx` — adds `'local'` to `useRemoteStt` so the mic renders on Firefox/Chromium. Also: `appendTextToDraft` now calls `focusPrompt()` after setting value (voice resolves async, focus had already left the textarea).

First call downloads the Whisper model into `~/.cache/huggingface` (~150 MB for `base`). Configurable via `stt.local.model` in `~/.hermes/config.yaml`.

## Test plan

- [ ] On Firefox over `localhost`, switch STT provider to Local, tap mic, speak — transcript lands in composer
- [ ] On Chromium (no Google SpeechRecognition backend), same path works
- [ ] Mid-recording, click off the composer, wait for transcript → focus returns to textarea automatically
- [ ] Confirm Groq and OpenAI paths are unchanged (no regression on cloud STT)
- [ ] First run triggers model download; subsequent runs are sub-second cold

### DIFF
--- a/scripts/transcribe-local.py
+++ b/scripts/transcribe-local.py
@@ -1,0 +1,93 @@
+#!/usr/bin/env python3
+"""Bridge script: transcribe an audio file using Hermes' built-in faster-whisper.
+
+Invoked by src/routes/api/transcribe.ts when stt.provider == "local". Reuses
+the Hermes _transcribe_local implementation so model caching, CUDA detection,
+CPU/int8 fallback, and ffmpeg conversion all match what the agent itself uses.
+
+Usage:
+  transcribe-local.py <audio_file> [--model NAME] [--language LANG]
+
+Prints a single JSON object to stdout:
+  {"ok": true,  "text": "..."}                     on success
+  {"ok": false, "error": "human-readable message"} on failure
+
+Always exits 0 — the JSON envelope carries the success/failure flag so the
+Node caller can parse a single contract without juggling exit codes.
+"""
+
+from __future__ import annotations
+
+import argparse
+import json
+import os
+import sys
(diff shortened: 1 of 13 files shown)


rmnelson:
Closing — opened without explicit per-PR approval from the contributor. Branch remains available at rmnelson:<branch>; will reopen if/when the change is ready to propose upstream.

PR #547, opened 2026-05-26
fix(avatar): graceful fallback when assistant avatar image is blocked

rmnelson:
## Summary

If \`/claude-avatar.webp\` fails to load — most often because Firefox strict tracking protection or an ad-blocker matches the path — the browser renders the alt text "Hermes Agent" inside the broken-image placeholder. At the 24px size used in chat messages, only "He" plus a clipped letter fits, leaving a confusing partial-word artifact under every assistant bubble.

Swap to an \`onError\`-tracked element. On load failure render an accent-tinted square with **H** instead of the broken-image alt-text placeholder. Accessibility label moves to \`aria-label\`/\`role\` on both the img and the fallback so screen readers still announce "Hermes Agent" without the visible alt text ever appearing.

## Test plan

- [ ] Block the image (Firefox: load with strict tracking protection on a path the filter list matches, OR open DevTools → Network → right-click \`claude-avatar.webp\` → **Block URL** → reload)
- [ ] Verify the fallback "H" square appears instead of clipped alt text
- [ ] Verify assistant avatar still renders normally when the image loads cleanly
- [ ] Verify screen-reader output is unchanged

### DIFF
--- a/src/components/agent-view/agent-view-panel.tsx
+++ b/src/components/agent-view/agent-view-panel.tsx
@@ -13,6 +13,7 @@ import {
   useReducedMotion,
 } from 'motion/react'
 import { AgentCard } from './agent-card'
+import { BackgroundRunsSection } from './background-runs-section'
 import { useAgentSpawn } from './hooks/use-agent-spawn'
 import type {
   AgentNode,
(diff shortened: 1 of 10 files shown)


rmnelson:
Closing — opened without explicit per-PR approval from the contributor. Branch remains available at rmnelson:<branch>; will reopen if/when the change is ready to propose upstream.

PR #577, opened 2026-05-31
fix(server): accept OpenAI-style list shape for gateway sessions/messages

morganjppeach:
## Problem

When the Hermes Agent gateway runs **without the dashboard** (enhanced-fork mode), `listSessions()` and `getMessages()` call the gateway's `/api/sessions` and `/api/sessions/:id/messages` endpoints. Those return an **OpenAI-style list** — `{ object: "list", data: [...] }` — not the `{ items: [...] }` shape the code assumed.

So `resp.items` was `undefined`, and:
- `/api/sessions` did `undefined.map(...)` → **HTTP 500** `Cannot read properties of undefined (reading 'map')`
- `/api/history` did `undefined.length` → **HTTP 500** `Cannot read properties of undefined (reading 'length')`

In the UI this surfaced as a persistent **"Connection error — Try refreshing or check Settings → Advanced → Hermes"** banner, even though chat itself worked fine (it doesn't depend on the session list).

## Fix

In `src/server/claude-api.ts`, both functions now accept either response shape and never return `undefined`:

```js
return resp.items ?? resp.data ?? []
```

## Testing

Against a live gateway in enhanced-fork mode (no dashboard):
- `GET /api/sessions` → was 500, now **200** with the session list
- `GET /api/history?...` → was 500, now **200** with mapped messages
- `/api/ping`, `/api/models`, `/api/session-status`, `/api/sessions/:id/status`, `/api/sessions/:id/active-run` → unaffected (still 200)

These were the only two `resp.items` sites in the file.

### DIFF
--- a/src/server/claude-api.ts
+++ b/src/server/claude-api.ts
@@ -133,10 +133,15 @@ export async function listSessions(
     const resp = await listDashboardSessions(limit, offset)
     return resp.sessions as Array<ClaudeSession>
   }
-  const resp = await claudeGet<{ items: Array<ClaudeSession>; total: number }>(
-    `/api/sessions?limit=${limit}&offset=${offset}`,
-  )
-  return resp.items
+  const resp = await claudeGet<{
+    items?: Array<ClaudeSession>
+    data?: Array<ClaudeSession>
+    total?: number
+  }>(`/api/sessions?limit=${limit}&offset=${offset}`)
+  // The gateway (OpenAI-compat) returns { object: 'list', data: [...] }, while the
+  // dashboard / older gateway shape uses { items: [...] }. Accept either, and never
+  // return undefined (callers .map over this).
+  return resp.items ?? resp.data ?? []
 }
 
 export async function getSession(sessionId: string): Promise<ClaudeSession> {
@@ -195,10 +200,15 @@ export async function getMessages(
     const resp = await getDashboardSessionMessages(sessionId)
     return resp.messages as Array<ClaudeMessage>
   }
-  const resp = await claudeGet<{ items: Array<ClaudeMessage>; total: number }>(
-    `/api/sessions/${sessionId}/messages`,
-  )
-  return resp.items
+  const resp = await claudeGet<{
+    items?: Array<ClaudeMessage>
+    data?: Array<ClaudeMessage>
+    total?: number
+  }>(`/api/sessions/${sessionId}/messages`)
+  // Gateway (OpenAI-compat) returns { object: 'list', data: [...] }; dashboard / older
+  // shape uses { items: [...] }. Accept either, and never return undefined (callers
+  // read .length / .map / .slice on this).
+  return resp.items ?? resp.data ?? []
 }
 
 export async function searchSessions(


maximeagnes-ship-it:
I can confirm this issue on Hostinger Docker Manager with image ghcr.io/outsourc-e/hermes-workspace:latest.

Environment:
- Hermes Workspace image: ghcr.io/outsourc-e/hermes-workspace:latest
- Hermes Agent backend reachable at http://hermes-agent:8642
- Backend auth works: /v1/models and /v1/responses return 200
- Codex OAuth backend works
- /api/sessions?limit=50&offset=0 returns HTTP 200

Observed response from Hermes Agent:
{
  "object": "list",
  "data": [...]
}

Workspace UI error:
- Failed to load sessions
- Connection error
- Cannot read properties of undefined (reading 'map')

Hostinger/Kodee confirmed the frontend expects:
{
  "sessions": [...]
}

So the robust fix should probably accept all three shapes:
const sessions = payload.sessions ?? payload.items ?? payload.data ?? [];

This is blocking the web chat UI even though the backend and model are correctly connected.

Josephur:
Reproducing #577 here on Hermes Workspace v2.3.0 + Hermes Agent on a native Linux install (not Docker). Same `Cannot read properties of undefined (reading 'map')` symptom, plus what looks like a closely-related second failure mode that may belong in this issue or its own.

## Environment

- Hermes Workspace: `outsourc-e/hermes-workspace` v2.3.0 (latest tag), running via `pnpm dev` from source on Linux
- Hermes Agent: latest, listening on `127.0.0.1:8642`
- `HERMES_API_TOKEN` set in workspace `.env`
- Not Docker, no dashboard on `:9119` (just the gateway on `:8642`)

## What breaks

Loading `http://127.0.0.1:3000/` shows two stacked errors:

**Error 1 — `Failed to load sessions` (auth)**

```
Failed to load sessions. Hermes Agent dashboard /api/sessions?limit=50&offset=0: 401 {"error":{"message":"Invalid API key","type":"invalid_request_error","code":"invalid_api_key"}}
```

Repro:
```
$ curl -s -o /dev/null -w "%{http_code}\n" http://127.0.0.1:3000/api/sessions
{"ok":false,"error":"Unauthorized"}
$ curl -s -o /dev/null -w "%{http_code}\n" http://127.0.0.1:8642/api/sessions
401
```

The Vite dev proxy (`vite.config.ts:526`, the `/api/claude-proxy` block and the surrounding `/api` proxies) forwards the request to `127.0.0.1:8642` but never adds `Authorization: Bearer ${HERMES_API_TOKEN}`. The browser bundle can't read `HERMES_API_TOKEN` either (no `VITE_` prefix), so the request goes out with no auth header at all.

**Error 2 — `Cannot read properties of undefined (reading 'map')` (this issue)**

`src/screens/dashboard/dashboard-screen.tsx` calls `normalizeDashboardSessionsPayload(data)` then renders `rawSessions.map(...)`. Because error 1 returns `{error: ...}` instead of `{sessions: [...]}`, `rawSessions` is `undefined` and `.map` throws. This is exactly the class your PR fixes — `resp.items ?? resp.data ?? []` — except on the Vite proxy path the response is an error envelope, not an OpenAI list. So even with #577 merged, error 2 will still fire on auth-failed responses unless the dashboard side also handles the `{error: ...}` envelope from the gateway.

## Why I'm commenting here

- Your PR fixes the `.map` crash when the gateway returns the wrong shape ✓
- It doesn't fix the underlying 401 on my setup, because my gateway has auth on and the Vite proxy never sends the token
- A user with the same symptom but on Docker (`#577`'s comment from `maximeagnes-ship-it` yesterday) may or may not also be hitting the proxy-auth issue — hard to tell without the same repro from them

## Suggestion

If maintainers want one PR to close both error paths on the dashboard, the fix needs two parts:
1. Land #577 as-is (response-shape tolerance).
2. Add a `proxy.on('proxyReq', ...)` hook to the `/api/*` proxy blocks in `vite.config.ts` that injects `Authorization: Bearer ${process.env.HERMES_API_TOKEN}` server-side. ~10 lines.

Filing a separate issue for (2) since it's an unrelated code path, just want to flag the dependency here first.

## Logs / traces

- `vite.config.ts` lines 525-545 — current proxy block, no `configure` auth hook
- `src/components/agent-view/agent-stream-panel.tsx:116` — the `throw new Error('Failed to load sessions')` site
- `src/screens/dashboard/dashboard-screen.tsx` — the `.map` crash site
- `src/lib/gateway-api.ts:206` — primary `/api/sessions` call site

Happy to test a fix branch if one gets cut.

maximeagnes-ship-it:
Merci pour l’analyse, ça correspond bien au symptôme que j’ai rencontré
côté Docker/Hostinger Workspace : dashboard qui plante avec Cannot read
properties of undefined (reading 'map'), alors que le backend Hermes Agent
semble répondre correctement sur certains endpoints.

Dans mon cas, il serait utile de vérifier si le Workspace transmet bien
HERMES_API_TOKEN au gateway Hermes Agent via le proxy. Je suis d’accord que
le fix devrait probablement couvrir les deux niveaux :
1. tolérance du front aux formats de réponse inattendus ou erreurs ;
2. injection serveur du header Authorization: Bearer ${HERMES_API_TOKEN}
dans les proxys /api.

Je peux tester une branche corrective sur mon environnement si besoin.

Maxime

outsourc-e:
This landed in main some time ago (the exact { items | data } fallback + comments are in src/server/claude-api.ts on main, and #679 extended it further with { messages: [...] }). Closing as already-shipped — thanks for the original fix!

PR #617, opened 2026-06-10
fix: fall back to API_SERVER_KEY from HERMES_HOME/.env for HERMES_API_TOKEN

### LINKED ISSUE #613: macOS GUI app cannot read HERMES_API_TOKEN — no file-based fallback
## Problem

When Hermes Agent gateway has API_SERVER_KEY set (mandatory in hermes-agent v0.16.0), the workspace must send it as HERMES_API_TOKEN. The workspace reads this from process.env.HERMES_API_TOKEN only.

On macOS, GUI apps launched via Finder/Dock do NOT inherit custom environment variables from the shell. launchctl setenv is unreliable. Modifying Info.plist LSEnvironment breaks the ad-hoc code signature (error -54).

Result: double-clicking the app in /Applications launches the UI, gateway/dashboard auto-start, but chat silently fails because the bearer token is never sent.

## Root cause

prod-server.cjs sets defaults for HERMES_API_URL and HERMES_DASHBOARD_URL but has no fallback for HERMES_API_TOKEN.

## Fix (trivial)

In prod-server.cjs, add a file-based fallback:

```js
if (!process.env.HERMES_API_TOKEN) {
  try {
    const envFile = require('path').join(require('os').homedir(), '.hermes', '.env')
    const envContent = require('fs').readFileSync(envFile, 'utf-8')
    const match = envContent.match(/^API_SERVER_KEY=(.+)$/m)
    if (match) process.env.HERMES_API_TOKEN=match[...)
  } catch (_) {}
}
```

Reads the same key the gateway already uses, from the same file the gateway already reads.

## Related

Issue #234 fixed the env var name but not the source problem for macOS.

GabrielBBaldez:
Fixes #613.

## What

GUI apps launched from Finder/Dock on macOS don't inherit shell environment variables, so `HERMES_API_TOKEN` is never set for the desktop workspace — chat silently fails against gateways that have `API_SERVER_KEY` enabled (mandatory since hermes-agent v0.16.0). `launchctl setenv` is unreliable and patching `Info.plist` breaks the ad-hoc signature, so there was no sane way to get the token into the app.

`electron/prod-server.cjs` already defaults `HERMES_API_URL` / `HERMES_DASHBOARD_URL`, but had no fallback for the token.

## Fix

Add `applyApiTokenFileFallback()` to `prod-server.cjs`: when `HERMES_API_TOKEN` is not set, read `API_SERVER_KEY` from `$HERMES_HOME/.env` (defaulting to `~/.hermes/.env`) — the same key, from the same file, the gateway itself reads.

Details:
- Runs **before** `loadServerBuild()`, because parts of the server bundle capture `HERMES_API_TOKEN` at module load time (e.g. the gateway bearer token in `gateway-capabilities.ts`).
- An explicitly set `HERMES_API_TOKEN` always wins — the fallback only fills the gap.
- Respects the `HERMES_HOME` override (same convention the rest of the codebase/tests use).
- Tolerates quotes/whitespace in the `.env` line; on any read/parse failure it falls through silently, leaving behavior exactly as it is today.

## Verification

- `node --check electron/prod-server.cjs` — clean.
- Exercised the exact shipped function against temp `HERMES_HOME` dirs: plain value, quoted value, surrounding whitespace, pre-set env var (not overridden), and missing file (no crash) — all behave as expected.
- `prettier --check` — the only remaining warning on the file is the pre-existing one present on `main` (line endings), unrelated to this change.

### DIFF
--- a/electron/prod-server.cjs
+++ b/electron/prod-server.cjs
@@ -1,5 +1,6 @@
 const http = require('http')
 const fs = require('fs')
+const os = require('os')
 const path = require('path')
 
 const portArg = process.argv.find(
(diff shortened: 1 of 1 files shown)


GabrielBBaldez:
@Interstellar-code friendly ping on this one — happy to rebase or adjust anything if that helps it move along.

PR #620, opened 2026-06-11
fix: remove --frozen-lockfile from Dockerfile to fix build failure

heuro-ai:


outsourc-e:
Superseded: #667 (merged today) fixes the underlying pnpm 10+ Docker failure by copying pnpm-workspace.yaml (carries allowBuilds approvals) into the build stage — keeping --frozen-lockfile so images stay reproducible. Removing frozen-lockfile would mask lockfile drift rather than fix it. Please retry on latest main and open an issue if the build still fails.

PR #636, opened 2026-06-16
fix: generate auto title for default "Local Chat" sessions

### LINKED ISSUE #635: Auto-session title not generated when fallback title is "Local Chat"
After commit 274e9a2e ("fix(chat): correct local session accounting and titles"), sessions without a title are displayed as "Local Chat" in the sidebar. However, the useAutoSessionTitle hook does not recognize "Local Chat" as a generic/placeholder title, so it never triggers automatic title generation for these sessions. Sessions remain permanently named "Local Chat" unless manually renamed.

Steps to reproduce:

Start a new chat session in the workspace (portable mode, local sessions)
Send a message and receive a response
Observe the session title in the sidebar
Expected behavior:
After the first assistant response, the session title should be automatically generated from the conversation content (e.g., "Discussion about project structure").

Actual behavior:
The session title stays as "Local Chat" indefinitely. No automatic title is ever generated.

Root cause:
In src/screens/chat/hooks/use-auto-session-title.ts, the GENERIC_TITLE_PATTERNS array contains patterns like "New Session", "Untitled", "Chat", etc., but does not include "Local Chat". The hook checks isGenericTitle() before generating a title — since "Local Chat" is not recognized as generic, shouldGenerate remains false.

The fallback string "Local Chat" was introduced in the session list API handler, but the auto-title hook was never updated to match.

Proposed fix:
Add "Local Chat" (and ideally a case-insensitive regex pattern) to GENERIC_TITLE_PATTERNS in use-auto-session-title.ts in:
```typescript
const GENERIC_TITLE_PATTERNS: (string | RegExp)[] = [
/^new\s*(session|chat)?$/i,
/^untitled$/i,
/^chat\s*\d*$/i,
/^local\s*chat$/i, // <-- add this
// ... existing patterns
];

### PR DESCRIPTION
**Pull request by JSap0914**
Fixes #635

## Bug
Sessions without a title fall back to the label `"Local Chat"` (set in `src/routes/api/sessions.ts`). The `useAutoSessionTitle` hook's `isGenericTitle` check did not recognize `"Local Chat"` as a generic/placeholder title, so `shouldGenerate` returned `false` and automatic title generation never ran. These sessions stayed permanently named "Local Chat" unless renamed by hand. (PR #350, which introduced the fallback label, did not add a matching pattern.)

## Fix
Add `/^local chat$/i` to `GENERIC_TITLE_PATTERNS` in `src/screens/chat/hooks/use-auto-session-title.ts`. Now `isGenericTitle('Local Chat')` is `true`, so the hook generates a title from the first user message after the first assistant response, like every other placeholder title.

`isGenericTitle` is now exported so it can be unit-tested directly.

## Verification
Added `src/screens/chat/hooks/use-auto-session-title.test.ts` covering the new case plus existing placeholder/real-title behavior.

```
npx vitest run src/screens/chat/hooks/use-auto-session-title.test.ts
# Test Files  1 passed (1)
#      Tests  3 passed (3)
```

Ran together with the nearest existing chat tests (`utils.test.ts`, `use-streaming-message.test.ts`): 9 passed.

### DIFF
--- a/src/screens/chat/hooks/use-auto-session-title.test.ts
+++ b/src/screens/chat/hooks/use-auto-session-title.test.ts
@@ -1,0 +1,23 @@
+import { describe, expect, it } from 'vitest'
+
+import { isGenericTitle } from './use-auto-session-title'
+
+describe('isGenericTitle', () => {
+  it('treats empty and placeholder titles as generic', () => {
+    expect(isGenericTitle('')).toBe(true)
+    expect(isGenericTitle('New Session')).toBe(true)
+    expect(isGenericTitle('Untitled')).toBe(true)
+    expect(isGenericTitle('Conversation')).toBe(true)
+  })
+
+  it('treats the default "Local Chat" fallback as generic (#635)', () => {
+    expect(isGenericTitle('Local Chat')).toBe(true)
+    expect(isGenericTitle('local chat')).toBe(true)
+    expect(isGenericTitle('  Local Chat  ')).toBe(true)
+  })
+
+  it('keeps real user-derived titles non-generic', () => {
+    expect(isGenericTitle('Refactor auth middleware')).toBe(false)
+    expect(isGenericTitle('Local development setup notes')).toBe(false)
+  })
+})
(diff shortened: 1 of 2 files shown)


Cephei-OpenSource:
Thank you!

PR #638, opened 2026-06-16
feat: translate agent bus panel from portuguese to spanish

infraitsalvita:
Translated all labels, descriptions, and status messages in the Agent Bus panel to Spanish to improve internationalization.


outsourc-e:
Closing alongside #646: the panel is now English in main (#639). If/when the app grows an i18n framework, a proper es-ES locale file would be welcome — hardcoded per-language forks of components aren't a pattern we can maintain. Thanks!

PR #646, opened 2026-06-19
i18n(agent-bus): translate Agent Bus panel from pt-BR to English

smatty-ice:
## What

The **Agent Bus operations panel** (`src/screens/agents/components/agent-bus-panel.tsx`, added in #477) has all of its UI strings hardcoded in Brazilian Portuguese. It is the only Portuguese-localized component under `src/` — every other screen in the app is in English — so it renders inconsistently for English users.

This PR translates all visible strings to English.

## Changes

- All visible UI text translated to English (headings, labels, empty states, button labels, status/loading/error messages).
- Date formatting locale changed `pt-BR` → `en-US` to match.
- **No behavior changes.** Props, state, fetch logic, action payloads, and target identifiers (`sync-roadmap`, `thumbnail-mission`, `handoff-mission`, and targets `vini` / `dona-helena` / `larissinha`) are all left exactly as-is — only the human-readable strings changed.

Diff is one file, 33 insertions / 33 deletions.

## Notes

Kept the scope strictly to translation so it stays trivially reviewable. If maintainers would prefer these strings routed through an i18n layer instead of hardcoded English, happy to follow up.

### DIFF
--- a/src/screens/agents/components/agent-bus-panel.tsx
+++ b/src/screens/agents/components/agent-bus-panel.tsx
@@ -54,14 +54,14 @@ type ActionState =
 
 const initialActionState: ActionState = {
   status: 'idle',
-  message: 'Acoes seguras ficam registradas no Agent Bus.',
+  message: 'Safe actions are logged in the Agent Bus.',
 }
 
 function formatDate(value?: string): string {
-  if (!value) return 'sem leitura'
+  if (!value) return 'no reading'
   const date = new Date(value)
   if (Number.isNaN(date.getTime())) return value
-  return new Intl.DateTimeFormat('pt-BR', {
+  return new Intl.DateTimeFormat('en-US', {
     day: '2-digit',
     month: '2-digit',
     hour: '2-digit',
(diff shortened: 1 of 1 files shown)


redx042:
Just a user here, not a contributor. Flagging that this bites real deployments: on an English instance the entire Operations → Agent Bus panel renders in Portuguese (from [#477](https://github.com/outsourc-e/hermes-workspace/pull/477)), so this fix is appreciated.

Heads-up in case it's useful: [#638](https://github.com/outsourc-e/hermes-workspace/pull/638) (→ Spanish) and [#639](https://github.com/outsourc-e/hermes-workspace/pull/639) (→ English) target the same file, so there's some overlap to sort out.

One thing I noticed that none of the three PRs seem to cover. The fix is strings-only, but the action buttons lower in the panel still call runAction() with hardcoded demo payloads from [#477](https://github.com/outsourc-e/hermes-workspace/pull/477) (sync-roadmap, "Missão Thumbnail Vini", "Handoff Helena para Larissinha" → Handoff Dona Helena -> Larissinha). They reference specific named agents and dispatch real actions, so they'd probably want handling in a follow-up. Thanks for taking this on!

outsourc-e:
Merged the equivalent earlier PR #639 (same file, same translation + en-US locale fix, submitted first). Thanks for the contribution regardless — appreciated!

PR #653, opened 2026-06-24
Claude/compassionate dirac vn6dl5

mvpbulliesllc-hash:


outsourc-e:
The content here looks genuinely useful (keys:check doctor + Vercel deploy path), but the PR has an auto-generated title and no description. Could you: (1) retitle to something like 'feat: API key doctor script + Vercel deployment support', (2) add a short description covering what keys:check probes and what the Vercel entry does, (3) confirm scripts/check-keys.mjs never prints secret values (fingerprint-only as the docs claim)? Happy to review properly once it's documented.

PR #656, opened 2026-06-25
fix(docker): build on pnpm 10+ — skip unneeded dep build scripts and drifted lockfile

### LINKED ISSUE #621: Published main-* images don't boot (regress #129); fresh-clone build also fails under pnpm 10/11
## Summary

Two separate, compounding problems make it currently impossible to run an up-to-date Hermes Workspace:

1. **The published CI images (`:latest`, `:main`, `main-<sha>`) do not boot** — they exit with code 0 in a restart loop. This is a regression of the *exact* bug that #129 fixed.
2. **Building from source on a fresh clone also fails** under current pnpm, so you can't even work around (1) by building it yourself without edits.

---

## Problem 1 — published images regress #129 (container exits code 0)

`ghcr.io/outsourc-e/hermes-workspace:latest` is **frozen at the 2026-05-08 build** (digest `sha256:2d2ba9aa5b12…`). It is the last image that actually boots. Every newer per-commit image, e.g. `:main-ffd6093` (`sha256:5c5058…`), starts and immediately exits 0, looping under any restart policy with **no log output**.

Root cause — the published image ships the wrong entrypoint:

```
$ docker image inspect ghcr.io/outsourc-e/hermes-workspace:main-ffd6093 \
    --format '{{json .Config.Cmd}}'
["node","--max-old-space-size=2048","dist/server/server.js"]

$ docker run --rm --entrypoint sh ghcr.io/outsourc-e/hermes-workspace:main-ffd6093 -c 'ls /app'
dist  node_modules  package.json  skills        # <- no server-entry.js, no .output/
```

`dist/server/server.js` is only the TanStack Start SSR fetch handler (`createServerEntry`, exports `server`); it never calls `.listen()`. So `node dist/server/server.js` runs top-level code and exits 0 — precisely what the Dockerfile comment warns about:

> `server-entry.js` is the Node HTTP server that wraps the TanStack Start fetch handler exported by `dist/server/server.js`. Without it, `node dist/server/server.js` imports the handler module, runs top-level code, and exits (code 0) … see issue #129.

But the **repo Dockerfile is correct** (`COPY … server-entry.js`, `CMD ["node","…","server-entry.js"]`). So the published images are **not built from the repo Dockerfile** (or from a stale one). The CI publish pipeline appears to use a different/older build whose CMD points at the non-listening module and omits `server-entry.js`. Net effect: `:latest` was never advanced past 2026-05-08 because newer builds don't run.

**Repro**
```
docker run --rm ghcr.io/outsourc-e/hermes-workspace:main-ffd6093   # exits 0, no output
docker run --rm ghcr.io/outsourc-e/hermes-workspace:latest         # works (2026-05-08 build)
```

**Fix:** make the CI publish job build from the in-repo `Dockerfile` (which already encodes the #129 fix), and re-point `:latest`/`:main` to it.

---

## Problem 2 — `pnpm install` fails on a fresh clone (can't self-build the workaround)

Building from source with the repo `Dockerfile` (`node:22-slim` + `corepack enable`) fails at `pnpm install`:

```
[ERR_PNPM_IGNORED_BUILDS] Ignored build scripts: electron-winstaller, electron, esbuild, unrs-resolver
Run "pnpm approve-builds" to pick which dependencies should be allowed to run scripts.
ERROR: process "… pnpm install --frozen-lockfile" did not complete successfully: exit code: 1
```

Two contributing causes:

- **No `packageManager` field** in `package.json`, so `corepack` pulls the latest pnpm (currently **11.5.3**), which *hard-errors* on un-approved dependency build scripts instead of warning.
- **`pnpm-workspace.yaml` uses an unrecognized key:**
  ```yaml
  allowBuilds:
    electron: true
    esbuild: true
    …
  ```
  `allowBuilds` is not a pnpm setting. The valid keys are `onlyBuiltDependencies` (allowlist) / `ignoredBuiltDependencies` (denylist). Because the allowlist never registers, every build script is "ignored-pending-approval" → the hard error. (Under pnpm 9 the same file instead fails with `packages field missing or empty`, since pnpm 9 treats `pnpm-workspace.yaml` purely as a monorepo definition.)

**Workaround that builds cleanly** (esbuild/unrs-resolver ship native binaries via optional platform deps, so their postinstalls aren't needed — matching the working 2026-05-08 image):
```dockerfile
RUN pnpm install --no-frozen-lockfile --ignore-scripts
```

**Suggested fix:** pin `"packageManager": "pnpm@<version>"` in `package.json` and replace the `allowBuilds` map in `pnpm-workspace.yaml` with a proper `onlyBuiltDependencies` list (and/or `--ignore-scripts` in the Dockerfile).

---

## Environment
- Host: Linux, Docker, deploying via Coolify (two-service compose: `nousresearch/hermes-agent` + this image).
- Working image: `:latest` @ `sha256:2d2ba9aa5b12…` (2026-05-08).
- Broken images: `:main-ffd6093` @ `sha256:5c5058…` and other `main-<sha>` builds.
- Source built & verified locally from `main` @ `d04e1f3` with the `--ignore-scripts` change — boots correctly (`Hermes Workspace running at http://0.0.0.0:3000`).

FelixBetzin:
## Problem
A fresh Docker build currently fails (#621). On the un-pinned pnpm, two issues compound:

1. corepack pulls the latest pnpm (11.x). pnpm 10+ aborts `pnpm install` with `ERR_PNPM_IGNORED_BUILDS` when dependencies ship build scripts that aren't explicitly approved (`electron`, `electron-winstaller`, `esbuild`, `unrs-resolver`).
2. `--frozen-lockfile` separately breaks fresh clones whenever the lockfile drifts (the motivation behind #620).

So the image never builds end-to-end on a current toolchain.

## Fix
Scope the fix to the image build, where **no dependency build scripts are actually needed**:

- `esbuild` ships its platform binary via `optionalDependencies` (no postinstall required),
- `electron` / `electron-winstaller` are desktop-only and unused in the web/server image.

`--ignore-scripts` is therefore safe here and keeps the web image lean (no pointless Electron download). `--no-frozen-lockfile` keeps fresh clones building when the lockfile drifts.

This complements/supersedes #620, which only drops `--frozen-lockfile`; that change alone still fails at `ERR_PNPM_IGNORED_BUILDS` on pnpm 10+.

## Verification
Builds and runs end-to-end (full `pnpm install` + `pnpm build` + container boot). The previous build failed exactly at:

```
[ERR_PNPM_IGNORED_BUILDS] Ignored build scripts: electron-winstaller@5.4.0, electron@40.9.3, esbuild@0.27.7, unrs-resolver@1.11.1
```

and succeeds with this change.

Fixes #621.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,7 +17,11 @@ WORKDIR /app
 
 # Install deps (cache-friendly: copy only manifests first)
 COPY package.json pnpm-lock.yaml* ./
-RUN pnpm install --frozen-lockfile
+# pnpm 10+ aborts install when deps have unapproved build scripts (ERR_PNPM_IGNORED_BUILDS),
+# and corepack pulls the latest pnpm. The web/server image needs none of those scripts:
+# esbuild ships prebuilt binaries via optionalDependencies and electron is desktop-only.
+# --no-frozen-lockfile also keeps fresh clones building when the lockfile drifts.
+RUN pnpm install --no-frozen-lockfile --ignore-scripts
 
 # Copy sources and build
 COPY . .


outsourc-e:
Thanks for this — the ERR_PNPM_IGNORED_BUILDS diagnosis was right. We just merged #667 which fixes the same failure by copying pnpm-workspace.yaml (which carries the allowBuilds approvals) into the image, keeping build-script approvals declarative instead of blanket --ignore-scripts, and keeping --frozen-lockfile so image builds stay reproducible. If pnpm 10/11 still fails for you on latest main after that, please open an issue with the exact error — the lockfile-drift half (#621) may still need attention separately.

PR #663, opened 2026-06-28
fix(gateway): honor HERMES_DASHBOARD_URL in dashboard auto-detect (cross-user session leak)

florian8211:
## What

`autoDetectDashboardUrl()` only skipped discovery when `CLAUDE_DASHBOARD_URL` was set, ignoring **`HERMES_DASHBOARD_URL`** — the documented primary env var (see the resolution order in the file header). When an operator sets `HERMES_DASHBOARD_URL` to a non-default port, the guard did not fire and the probe hit the hard-coded `:9119` candidate. If any dashboard answered there, it silently overwrote the explicit URL.

The sibling `autoDetectGatewayUrl()` already guards on **both** `HERMES_API_URL` and `CLAUDE_API_URL`. This one-line fix mirrors that so an explicit setting always wins.

## Why it matters (real-world bug)

Multi-user setup on one host — two users, each running their own gateway + dashboard on distinct ports:

| User | Gateway | Dashboard |
|------|---------|-----------|
| A    | `:8642` | `:9119`   |
| B    | `:8643` | `:9120`   |

User B's workspace was correctly configured with `HERMES_API_URL=:8643` and `HERMES_DASHBOARD_URL=:9120`. Chat and `/api/ping` correctly reported `:8643`. **But the session list showed user A's sessions.**

Root cause: `autoDetectDashboardUrl()` didn't see `HERMES_DASHBOARD_URL`, ran discovery, hit the hard-coded `:9119` candidate, A's dashboard answered `200`, and B's `CLAUDE_DASHBOARD_URL` got overwritten with A's port. Result: B's workspace rendered A's full session list (titles + previews) — a **cross-user privacy leak** — despite a correct explicit config.

Confirmed empirically: B's workspace node process opened a live TCP connection to `:9119` (A's dashboard) during session listing; after the fix it connects only to B's own `:8643`/`:9120`, and the sidebar shows only B's own sessions.

## Change

```diff
 async function autoDetectDashboardUrl(): Promise<void> {
-  if (process.env.CLAUDE_DASHBOARD_URL) return
+  if (process.env.HERMES_DASHBOARD_URL || process.env.CLAUDE_DASHBOARD_URL) return
```
(plus an explanatory comment)

## Tests

Added a regression test in `src/server/__tests__/gateway-capabilities.test.ts` asserting that when `HERMES_DASHBOARD_URL` is set:
- the `:9119` auto-detect probe never runs, and
- the explicit URL is preserved.

Verified the test **fails on the unpatched code** and passes with the fix. The `gateway-capabilities` suite is fully green (15/15). (Unrelated pre-existing failures in `mcp-presets-store`, `mcp-hub-sources-store`, `i18n`, and the Playwright `e2e/` specs exist on `main` and are untouched by this change.)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,15619 @@
+{
+  "name": "hermes-workspace",
+  "version": "2.3.0",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {
+    "": {
+      "name": "hermes-workspace",
+      "version": "2.3.0",
+      "license": "MIT",
+      "dependencies": {
+        "@base-ui/react": "^1.1.0",
+        "@hugeicons/core-free-icons": "^3.1.1",
+        "@hugeicons/react": "^1.1.4",
+        "@lobehub/icons": "^5.0.1",
+        "@lobehub/icons-static-png": "^1.83.0",
+        "@monaco-editor/react": "^4.7.0",
+        "@react-three/drei": "^10.7.7",
+        "@react-three/fiber": "^9.6.1",
+        "@react-three/postprocessing": "^3.0.4",
+        "@react-three/rapier": "^2.2.0",
+        "@tailwindcss/vite": "^4.1.18",
+        "@tanstack/react-query": "5.90.21",
+        "@tanstack/react-router": "1.166.7",
+        "@tanstack/react-router-devtools": "1.166.7",
+        "@tanstack/react-router-ssr-query": "1.166.7",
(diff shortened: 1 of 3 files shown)


florian8211:
Closing in favor of a clean branch — this one accidentally included an unrelated 15k-line package-lock.json from a local auto-sync commit. Reopening from a branch based directly on upstream/main with only the 2-file fix.

PR #682, opened 2026-07-03
fix(workspace): make fetchDashboardToken resilient to dashboard 500

xdemocle:
## Problem

When the Hermes Agent dashboard requires auth (302 → `/auth/login`) and the login page is broken (500), `fetchDashboardToken` throws `Dashboard index failed: 500`. This propagated to the dashboard overview page rendering a hard 500 error instead of gracefully degrading.

## Root Cause

`fetchDashboardToken` in `src/server/gateway-capabilities.ts` fetches the dashboard root URL to scrape the ephemeral session token. When dashboard auth is enabled (redirects `/` → `/auth/login`), and the login page itself returns 500, the token extraction fails with an unhandled throw.

## Fix

Instead of throwing on failure, return an empty string and log a `console.warn`. The callers (`dashboardFetch` → `safeJson`) already handle non-ok/401 responses gracefully by returning `null` per-section. This means the dashboard overview degrades to null sections instead of a hard 500 page.

## Re-application

A local patch is saved at `~/.hermes/patches/hermes-workspace-dashboard-token-500.patch` — can be re-applied with:

```sh
cd ~/hermes-workspace && git apply ~/.hermes/patches/hermes-workspace-dashboard-token-500.patch
```

This is useful after hermes-agent updates fix the underlying dashboard login 500.

### DIFF
--- a/src/server/__tests__/gateway-capabilities.test.ts
+++ b/src/server/__tests__/gateway-capabilities.test.ts
@@ -173,6 +173,23 @@ describe('gateway-capabilities', () => {
       await expect(mod.fetchDashboardToken()).resolves.toBe('live-token')
       expect(fetchMock.mock.calls.some(([url]) => url === 'http://127.0.0.1:9119/')).toBe(true)
     })
+
+    it('returns an empty token instead of throwing when dashboard root fails', async () => {
+      fetchMock.mockResolvedValue({
+        ok: false,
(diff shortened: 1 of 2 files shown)


xdemocle:
Closing as superseded by #683 (which includes this commit plus the cookie-jar login fix).

PR #661, opened 2026-06-28
fix(gateway): add cookie-based dashboard auth for agent v0.17.0+

tonyskifo:
## Problem

Hermes Agent v0.17.0 hardened dashboard authentication.

`HERMES_DASHBOARD_INSECURE=1` is now a no-op. Unauthenticated dashboard routes redirect to `/login`, and the dashboard middleware rejects `Authorization: Bearer` headers in favour of the `hermes_session_at` session cookie.

The workspace currently relies on scraping `window.__HERMES_SESSION_TOKEN__` from the unauthenticated dashboard HTML. Since that token is no longer exposed before authentication, all dashboard API calls fail with 401 (`no_cookie`) against Agent v0.17.0+.

Closes #650.

---

## Solution

This PR adds a cookie-authentication path for newer Hermes Agents.

When the following environment variables are configured:

- `HERMES_DASHBOARD_BASIC_AUTH_USERNAME`
- `HERMES_DASHBOARD_BASIC_AUTH_PASSWORD`

the workspace now:

1. Authenticates via `POST /auth/password-login`
2. Extracts the `hermes_session_at` cookie
3. Caches the cookie until shortly before expiry
4. Sends the cookie on dashboard requests instead of a Bearer token
5. Clears the cookie cache automatically after a 401 and retries

Deployments that do not configure these variables continue using the existing HTML token scraping behaviour.

---

## Build fix

Also includes a small Docker build improvement:

- copy `pnpm-workspace.yaml` before `pnpm install`
- exclude `.pnpm-store` from the Docker build context

This fixes `ERR_PNPM_IGNORED_BUILDS` during Docker builds using pnpm 11.

---

## Testing

Tested successfully against Hermes Agent v0.17.0.

- Dashboard login returns HTTP 200
- `hermes_session_at` cookie is issued
- Dashboard API requests succeed using cookie authentication
- Workspace health endpoint remains healthy
- No authentication warnings appear in workspace logs

---

## Risk

Low.

The new authentication path is only used when both dashboard credentials are configured.

Existing deployments continue using the legacy authentication flow unchanged.

PR #683, opened 2026-07-03
feat(workspace): POST-login to dashboard and reuse session cookie

xdemocle:
## Problem

When the workspace dashboard is configured with `basic_auth`, the workspace's `fetchDashboardToken` scrapes the root URL `'/'` for an inline session token. With auth enabled the dashboard redirects to `/auth/login` (or returns 500 from the OAuth `NotImplementedError` path), so the scrape returns empty and every protected `/api/*` call ends up with `401 {"reason":"no_cookie"}`.

A previous fix made the scrape path `throw new Error("Dashboard index failed: 500")`, which produced the user-visible `Dashboard index failed: 500` symptom in `/api/dashboard/overview` etc.

## Fix

Two commits, intentionally separable:

### Commit 1 — `fix(workspace): make fetchDashboardToken resilient to dashboard 500`

The inner scrape no longer throws on 500; it `console.warn`s and returns `''` so the caller can degrade gracefully (the existing `safeJson` path already handles 401/non-ok). This is identical to the diff in [#682](https://github.com/outsourc-e/hermes-workspace/pull/682) and is included here so this PR is reviewable against `main` in isolation. (#682 can be closed once this lands.)

### Commit 2 — `feat(workspace): POST-login to dashboard and reuse session cookie`

Adds the real fix for the underlying 401:

1. After the root scrape returns empty, `POST` credentials to `/auth/password-login`.
2. Capture `Set-Cookie` into a module-scoped jar (`dashboardCookieJar`).
3. `dashboardAuthHeaders` now returns `{ Cookie: <jar> }` instead of `{ Authorization: Bearer <token> }` — the dashboard validates `/api/*` via session cookie, not bearer.
4. `dashboardFetch` adds `credentials: 'include'` and clears the jar on 401 so the next attempt re-authenticates.
5. On 401 the cached jar is wiped so the retry actually re-logs in (instead of replaying the now-revoked session).

Credentials are read from `HERMES_DASHBOARD_BASIC_AUTH_USERNAME` / `HERMES_DASHBOARD_BASIC_AUTH_PASSWORD` (with `CLAUDE_DASHBOARD_*` and `HERMES_DASHBOARD_*` fallbacks for naming parity across Hermes + Claude codename versions).

## Tests

`pnpm exec vitest run src/server/__tests__/gateway-capabilities.test.ts` → **17/17 passing.**

- `returns an empty token instead of throwing when dashboard root fails` (commit 1)
- `falls back to /auth/password-login when root scrape returns 500` (commit 2 — verifies POST to `/auth/password-login`, cookie capture, and `dashboardAuthHeaders` returning `{ Cookie: session=abc123 }`)
- `returns empty token and warns when password login fails` (commit 2 — verifies 401 from `/auth/password-login` is surfaced and `dashboardAuthHeaders` returns `{}`)

Pre-existing failures in unrelated test files (e.g. `chat-message-list.test.tsx`) exist on `origin/main` too — not introduced here.

## Relation to #682

#682 contains only commit 1 (the resilience fix). This PR also contains commit 1 because the cookie-jar fallback in commit 2 needs the inner scrape to return `''` instead of throwing — they're operationally a pair. After this PR lands, #682 should be closed as superseded. The two commits are kept as separate commits in the branch history so reviewers can see them in isolation.

### DIFF
--- a/src/server/__tests__/gateway-capabilities.test.ts
+++ b/src/server/__tests__/gateway-capabilities.test.ts
@@ -173,6 +173,23 @@ describe('gateway-capabilities', () => {
       await expect(mod.fetchDashboardToken()).resolves.toBe('live-token')
       expect(fetchMock.mock.calls.some(([url]) => url === 'http://127.0.0.1:9119/')).toBe(true)
     })
+
+    it('returns an empty token instead of throwing when dashboard root fails', async () => {
+      fetchMock.mockResolvedValue({
+        ok: false,
+        status: 500,
+        text: async () => 'Internal Server Error',
+      })
+
+      const warnSpy = vi.spyOn(console, 'warn').mockImplementation(() => {})
+      const mod = await loadMod()
+
+      await expect(mod.fetchDashboardToken()).resolves.toBe('')
+      expect(warnSpy).toHaveBeenCalledWith(
+        '[gateway] Dashboard index returned 500 — token unavailable',
+      )
+      warnSpy.mockRestore()
+    })
   })
 
   it('does not mark Conductor available when dashboard returns SPA HTML fallback', async () => {
--- a/src/server/__tests__/gateway-capabilities.test.ts
+++ b/src/server/__tests__/gateway-capabilities.test.ts
@@ -190,6 +190,109 @@ describe('gateway-capabilities', () => {
       )
       warnSpy.mockRestore()
     })
+
+    it('falls back to /auth/password-login when root scrape returns 500', async () => {
+      process.env.HERMES_DASHBOARD_BASIC_AUTH_USERNAME = 'hermes'
+      process.env.HERMES_DASHBOARD_BASIC_AUTH_PASSWORD = 'qwe123'
+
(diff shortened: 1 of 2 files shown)


xdemocle:
Closing in favor of #661 — that PR implements the same cookie-based dashboard auth (POST /auth/password-login → Cookie: hermes_session_at=...) with a cleaner approach (explicit DashboardAuthMode, JWT caching with auto-refresh, Docker bonus fix). It's also older (Jun 28) so it should land first.

The unique value here — console.warn + return '' instead of throw new Error() when the dashboard root URL returns 500 — has been split out into #686 (fix(workspace): make fetchDashboardToken resilient to dashboard 500). Once #661 lands, the resilience change can be re-evaluated; if it's still useful, #686 is independent and small enough to merge separately.

Thanks to @tonyskifo for the cleaner take. Closing to avoid merge conflicts and duplicate review work.

xdemocle:
Abandoning hermes-workspace — too buggy. Local fix preserved at ~/.hermes/patches/hermes-workspace-dashboard-cookie-login.patch.

PR #684, opened 2026-07-04
fix(workspace): emit Content-Security-Policy as HTTP header (survives edge body rewrites)

xdemocle:
## Problem

When Cloudflare's JS Challenge (Bot Management → Security → JavaScript
Challenges, "Browser Integrity Check") trips on a request to the
workspace, Cloudflare injects an interstitial HTML body that includes
its own `<meta http-equiv="Content-Security-Policy">` tag with
per-request nonces and stricter directives. After the JS Challenge
passes, the browser executes that body, which then redirects to the
real page. The redirected real page's CSP was being delivered only
as a `<meta>` tag in `src/routes/__root.tsx` — and even that was
being **replaced** by Cloudflare's injected `<meta>` on first
challenge render.

This produced two user-visible bugs on `https://hw-srv3.rocco.me/`:

1. Browser console flooded with:
   - `The source list for the Content Security Policy directive 'default-src' contains an invalid source: '''. It will be ignored.`
   - `Loading the script 'https://hw-srv3.rocco.me/assets/main-...js' violates the following Content Security Policy directive: "default-src ' self'"... The action has been blocked.`
   - All hashed assets (JS, CSS, images, manifests) blocked.

2. Cloudflare Insights beacon (`static.cloudflareinsights.com/beacon.min.js`)
   was also blocked, breaking Cloudflare analytics.

## Fix

Move the CSP from a `<meta>` tag in the SPA shell to an **HTTP
response header** emitted by `server-entry.js` on every response
(including 404s, asset routes, and SSR responses). HTTP headers
survive edge body transformations — Cloudflare's JS Challenge
replaces the *body* but not the *response headers*, so the policy
the browser actually applies is ours, not theirs.

### What changes

- **New `src/lib/csp.ts`** — single source of truth for the CSP
  directives. Exports `APP_CSP` (used by `__root.tsx` as a `<meta>`
  fallback for offline/SSR contexts) and `APP_CSP_HEADERS` (joined
  string used in HTTP headers).
- **Updated `src/routes/__root.tsx`** — emits `<meta http-equiv
  "Content-Security-Policy">` from `APP_CSP`, but only when the
  same policy is not already on the response (which it always is
  in production via the HTTP header).
- **Updated `server-entry.js`** — adds `Content-Security-Policy`
  to `ALWAYS_HEADERS` and merges it on top of SSR-emitted headers
  in `requestHandler`. `ALWAYS_HEADERS` also includes
  `X-Content-Type-Options: nosniff` and `Referrer-Policy:
  strict-origin-when-cross-origin` for defense-in-depth.
- **Updated `vite.config.ts`** — adds the same CSP HTTP header in
  dev/preview modes via Vite middleware.
- **New `src/lib/csp.test.ts`** — 5 unit tests covering directive
  composition, escaping, and parity between `APP_CSP` and
  `APP_CSP_HEADERS`.

### Why this is the right fix

- HTTP headers are **not** subject to edge body rewrites, only `<meta>`
  CSPs are.
- All assets, inline scripts, and inline styles are now allowed
  (`'unsafe-inline'` for both `script-src` and `style-src`). Tightening
  to nonce-based is the next-step work — out of scope here. The dash
  comment in `server-entry.js` flags this for follow-up.
- Cloudflare Insights beacon is now allow-listed via `https:` in
  `img-src` and `script-src` (already the case via `https://cdn.jsdelivr.net`
  allowing JS, plus the inline `script-src 'unsafe-inline'` for the
  beacon boot). If a stricter CSP is needed later, this can be
  re-tightened to `https://static.cloudflareinsights.com` explicitly.

## Test plan

```
pnpm exec vitest run src/lib/csp.test.ts
```
→ 5/5 passing.

Manual smoke on `https://hw-srv3.rocco.me/`:
- Browser console clean (no CSP errors).
- All assets load (JS bundles, CSS, fonts, images).
- Inline scripts (`APP_INITIAL_STATE` hydration, `setup-step-content`)
  execute.
- Manifest.json and `apple-mobile-web-app-capable` deprecation
  warnings remain (cosmetic, separate issue).

Verified via tunnel probe:
```
HTTP/2 200
content-security-policy: default-src 'self'; base-uri 'self'; object-src 'none';
  form-action 'self'; script-src 'self' 'unsafe-inline' https://cdn.jsdelivr.net;
  style-src 'self' 'unsafe-inline' https://fonts.googleapis.com https://cdn.jsdelivr.net;
  img-src 'self' data: blob: https:; font-src 'self' data: https://fonts.gstatic.com;
  connect-src 'self' ws: wss: http: https:; worker-src 'self' blob:;
  media-src 'self' blob: data:; frame-src 'self' http: https:
```

## Notes for reviewers

- Keep `src/lib/csp.ts` in sync if you change `APP_CSP` in
  `__root.tsx`. Both come from the same module now.
- The `Content-Security-Policy` HTTP header is now **mandatory** —
  the `<meta>` in `__root.tsx` is a redundant fallback. If you'd
  rather remove the `<meta>` entirely for cleanliness, that's a
  follow-up PR.

Refs: hw-srv3.rocco.me CSP errors reported 2026-07-04

### DIFF
--- a/server-entry.js
+++ b/server-entry.js
@@ -7,6 +7,38 @@ import server from './dist/server/server.js'
 const __dirname = fileURLToPath(new URL('.', import.meta.url))
 const CLIENT_DIR = join(__dirname, 'dist', 'client')
 
+// Content Security Policy — emitted as an HTTP response header on EVERY
+// response so the policy survives any edge body transformations (e.g.
+// Cloudflare's JS Challenge inserting a `<meta http-equiv="Content-Security-Policy">`
+// with a per-request nonce into the served HTML when a request trips the
+// "impersonate browsers" rule).
+//
+// KEEP IN SYNC with `src/lib/csp.ts` (single source of truth) and
+// `src/routes/__root.tsx` APP_CSP (which emits the same policy as `<meta>`).
+// If you change the directives here, change them in all three places.
+const APP_CSP_HEADERS = [
+  "default-src 'self'",
+  "base-uri 'self'",
+  "object-src 'none'",
+  "form-action 'self'",
+  "script-src 'self' 'unsafe-inline' https://cdn.jsdelivr.net",
+  "style-src 'self' 'unsafe-inline' https://fonts.googleapis.com https://cdn.jsdelivr.net",
+  "img-src 'self' data: blob: https:",
+  "font-src 'self' data: https://fonts.gstatic.com",
+  "connect-src 'self' ws: wss: http: https:",
+  "worker-src 'self' blob:",
+  "media-src 'self' blob: data:",
+  "frame-src 'self' http: https:",
+].join('; ')
+
+const ALWAYS_HEADERS = {
+  'Content-Security-Policy': APP_CSP_HEADERS,
+  // Tighten later if/when CSP moves to nonce-based — the dash prefix
+  // makes adding/removing trivial without searching the codebase.
+  'X-Content-Type-Options': 'nosniff',
+  'Referrer-Policy': 'strict-origin-when-cross-origin',
+}
+
 const port = parseInt(process.env.PORT || '3000', 10)
 // Default HOST to localhost-only. Operators who want the workspace reachable
 // on a LAN / Tailscale / public surface must opt in explicitly with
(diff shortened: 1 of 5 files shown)


outsourc-e:
Superseded by #685, which contained this exact CSP-header commit plus the no-store /api/* hardening — both are now in main. Thanks @xdemocle, the edge-body-rewrite rationale was a great catch.

PR #688, opened 2026-07-05
[codex] Add Hermes runs approval flow

gos01363:
## Summary

Adds a Hermes Runs API transport for Workspace chat so tool approval requests can surface as approval cards and be resolved from the Workspace UI.

## Why

The OpenAI-compatible chat-completions path can return an approval-required error, but it does not provide the interactive approval event and approval response flow that Hermes exposes through `/v1/runs`.

## Changes

- Start portable Hermes chat requests through `POST /v1/runs` when available.
- Stream run events from `/v1/runs/{run_id}/events`.
- Convert `approval.request` events into Workspace approval UI events.
- Add approve/deny routes that call `POST /v1/runs/{run_id}/approval`.
- Keep a fallback to the existing path when a run cannot be started.
- Add focused unit coverage for the Runs API helper.

## Validation

- `pnpm exec vitest run src/server/hermes-runs-api.test.ts`
- `pnpm run build`
- Manual local validation with `approvals.mode: manual`: an approval card appeared in Workspace and approval completed the blocked run.

## Attribution

Implementation and validation were assisted by OpenAI Codex.


gos01363:
Closing for now; approval prompt still does not appear reliably. We'll reopen/push after the fix is verified.

PR #713, opened 2026-07-10
fix(sessions): fall back to gateway when dashboard create is unavailable

alexgunsberg:
## Summary

When Workspace detects a dashboard but its session-create endpoint is not implemented (404/405/501), fall back to the gateway `POST /api/sessions` path instead of making session creation fail. Other dashboard errors still propagate.

This keeps split-host and partially wired dashboard deployments usable while preserving strict failure handling for unexpected errors.

## Verification

- `git diff --check origin/main...HEAD`
- `pnpm run build`
- `pnpm exec vitest run src/server/__tests__/gateway-capabilities.test.ts` (16 passed)

No live service or active checkout was changed.

### DIFF
--- a/src/server/claude-api.ts
+++ b/src/server/claude-api.ts
@@ -154,18 +154,33 @@ export async function getSession(sessionId: string): Promise<ClaudeSession> {
   return resp.session
 }
 
+function shouldFallbackFromDashboardCreateSession(error: unknown): boolean {
+  const message = error instanceof Error ? error.message : String(error)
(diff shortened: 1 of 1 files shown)


alexgunsberg:
Closing under Alex's zero-review autonomous delivery policy. Upstream grants this account read-only permission, so the PR cannot be auto-merged. The exact head remains preserved on the alexgunsberg fork; no review action is requested from Alex.

PR #729, opened 2026-07-28
Trading engine: audit-visibility fix, grid stop-floor, Postgres test-isolation fix, taker-imbalance research

NPFernando:
## Summary

Eight commits from a 2026-07-27/28 investigation session, cleanly split off `feat/task-blocker-system` (which bundles unrelated task-blocker/approval-system work) onto its own branch for review.

- **Audit-visibility fix**: `demo-trading-engine.ts`'s `bail()` (kill switch / connectivity breaker / wrong trading mode gates) previously returned with zero audit trail — a tripped gate was indistinguishable from the engine not running at all. This is what turned a 2026-07-25 connectivity-breaker trip into what looked like days of total silence. Every bail now writes a `demo_trading_cycle_bailed` audit entry.
- **Grid absolute stop floor**: closes a real gap where `autoRecenter`'s stop-triggered re-arm recomputes the range from the current window, so the effective stop slides down with a falling market instead of acting as a hard floor. Backtest-validated (8-fold/4-year walk-forward: -104.82 → -97.64 USDT, PF 0.89 → 0.90, only fold 5 — the historically bad one — changed) and ported to the live incremental engine. **Off by default**, not armed.
- **Finance Postgres test-isolation fix**: `writeFinanceStore()` unconditionally mirrored to Postgres with no test-environment guard, so any `vitest run` could silently overwrite the real production Postgres finance database. Confirmed via nightly backups that `settings.demoTradingGrid` held test-fixture values for 2+ days with zero audit trail. Fixed at the single gate every read/write path bottlenecks through.
- **`taker_imbalance` strategy**: new council signal using Binance's `takerBuyBaseAssetVolume` kline field (already returned on every candle, no new data collection needed) as a buy/sell-pressure proxy. Caught and fixed a dead confidence formula before trusting the first backtest (same bug class as `sma_crossover`'s original issue — 0 trades until recalibrated). Real result after the fix: 554 trades, PF 0.97, clean NO-GO. Added to `STRATEGIES` only, never armed.

## Known pre-existing issue found, NOT fixed here (out of scope)

`origin/main` is currently broken for a fresh checkout: `research-store.ts` (existing since 2026-07-09, untouched by this PR) imports the `pg` npm package, which is not declared in `package.json`. This is the same root cause as a 2026-07-24 production crash-loop (`Cannot find package 'pg'`) — confirmed via `git show origin/main:package.json` (no `pg` entry) and a fresh `pnpm install` + `vitest run` in an isolated worktree reproducing the exact same error. Flagging for a separate fix; did not touch `package.json` in this PR to keep scope clean.

## Test plan
- [x] `npx tsc --noEmit` clean on all touched files (2 pre-existing unrelated errors elsewhere: the `pg` gap above, and an unrelated `tasks-replan-stubs.ts` type error)
- [x] `npx eslint` clean on all touched files
- [x] Full affected test suite: 224 tests pass across `trading-strategies`, `trading-backtest`, `demo-trading-engine`, `grid-backtest`, `grid-paper-engine`, `finance-postgres-store`, `binance-demo-client` (`llm-signal-engine.test.ts` fails to load due to the pre-existing `pg` gap above, unrelated to this PR's changes)
- [x] Verified cherry-picked cleanly onto `origin/main` with no conflicts, in an isolated git worktree


NPFernando:
Opened against the wrong repo by mistake (gh defaulted to the fork's upstream parent) — closing immediately, no review needed here. Correct PR going to NPFernando/hermes-workspace instead.

PR #758, opened 2026-08-17
feat: support Hermes Agent's dashboard login-form auth gate

lordzeezus:
## Summary

Hermes Agent's dashboard, once it has `HERMES_DASHBOARD_BASIC_AUTH_USERNAME`/`_PASSWORD` configured (as recommended for any non-loopback bind — see the security note in the Docker docs about the June 2026 unauthenticated-dashboard exploit), is **not** gated by RFC 7617 HTTP Basic Auth despite the "basic auth" naming. It's a cookie-session login form: `GET /` redirects to `/login`, and a real `Authorization: Basic ...` header is silently ignored (confirmed live). Auth actually happens via `POST /auth/password-login` with a `{provider,username,password,next}` JSON body, returning session cookies (`hermes_session_at` etc.) that must be replayed on subsequent requests.

`hermes-workspace`'s only mechanism for reaching the dashboard (`fetchDashboardToken()` in `gateway-capabilities.ts`) scrapes an ephemeral token from the dashboard's *unauthenticated* root-page HTML — which only works when the dashboard has no login gate. Once one is configured, that scrape gets a login redirect instead of a token, and there was no code path anywhere in this app for supplying credentials. The practical effect: every dashboard-backed feature (sessions list, skills, MCP, kanban, cron jobs, model info, config) is **permanently broken — not intermittently** — the moment an operator follows the Agent's own documented security recommendation for a non-loopback dashboard.

## Fix

Adds an opt-in login-and-cookie-cache path:
- New env vars `HERMES_DASHBOARD_BASIC_AUTH_USERNAME`/`_PASSWORD` — same names as the Agent side, so operators can copy the pair straight across.
- `loginToDashboard()`: POSTs to `/auth/password-login`, caches the resulting session cookie in memory (same in-flight-promise dedup pattern as the existing token cache), replays it as a `Cookie` header.
- `dashboardAuthHeaders()` tries the login-cookie path first when credentials are configured, falling back to the original token-scrape path otherwise — so deployments without a dashboard login gate are completely unaffected (byte-for-byte identical behavior).
- `dashboardFetch()`'s existing 401-retry-once logic now also clears the cookie cache, so a rotated/expired session re-logs in automatically.

## Test plan

- [x] 4 new tests in `gateway-capabilities.test.ts` (login + cache reuse, credentials sent correctly, re-login after 401, no-op when credentials aren't configured) — 20/20 passing in the file, stable across 5 repeated runs
- [x] Full suite regression check — no new failures introduced
- [x] `npx tsc --noEmit` — no new type errors (pre-existing unrelated errors elsewhere in the repo, none in the changed file)
- [x] Live-verified against a real deployment with the dashboard's login gate enabled: authenticated `GET /api/sessions` now returns `200` with the real session list instead of `401`

One implementation note for reviewers: this module has a fire-and-forget `void ensureGatewayProbed()` at the bottom that runs a capability probe on every import. In tests that reset modules per-test, that background probe races with whatever the test does next and independently exercises this same login path — cost some debugging time (initially looked like a caching bug) before landing on `await mod.ensureGatewayProbed()` + `fetchMock.mockClear()` before asserting, to isolate what's actually under test.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -120,6 +120,18 @@
 # Workspace scrapes the dashboard's ephemeral session token from the root HTML
 # automatically. Do not copy this token into .env: it changes whenever the
 # dashboard restarts and stale values cause 401s on /api/sessions and related APIs.
+# This only works when the dashboard has no login gate (loopback / same-host).
+
+# Dashboard login credentials (required if the Hermes Agent dashboard has
+# HERMES_DASHBOARD_BASIC_AUTH_USERNAME/_PASSWORD set on its own side, e.g. any
+# non-loopback deployment, where the auth gate is mandatory)
+#
+# Same variable names as the Agent side, so you can copy the pair straight
+# across. Without these, /api/sessions and other dashboard-backed APIs
+# (skills, MCP, cron jobs, model info) 401 the moment the dashboard requires
+# a login instead of serving its root page unauthenticated.
+# HERMES_DASHBOARD_BASIC_AUTH_USERNAME=
+# HERMES_DASHBOARD_BASIC_AUTH_PASSWORD=
 
 # Bypass fail-closed startup guard (NOT recommended)
 #
--- a/src/server/__tests__/gateway-capabilities.test.ts
+++ b/src/server/__tests__/gateway-capabilities.test.ts
@@ -42,6 +42,8 @@ beforeEach(() => {
   delete process.env.HERMES_DASHBOARD_URL
   delete process.env.HERMES_DASHBOARD_TOKEN
   delete process.env.CLAUDE_DASHBOARD_TOKEN
+  delete process.env.HERMES_DASHBOARD_BASIC_AUTH_USERNAME
+  delete process.env.HERMES_DASHBOARD_BASIC_AUTH_PASSWORD
   delete process.env.HOST
 })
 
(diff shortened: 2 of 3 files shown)


LugMuad:
Found one remaining auth issue with the Kanban proxy.

`src/server/kanban-dashboard-proxy.ts` still uses `fetchDashboardToken()`, so when the Dashboard is protected by Basic Auth it doesn't send the session cookie created by the new login flow in #758.

I changed it to use the new `dashboardAuthHeaders()` helper instead:

```diff
 import {
   CLAUDE_DASHBOARD_URL,
-  fetchDashboardToken,
+  dashboardAuthHeaders,
 } from './gateway-capabilities'

 async function buildHeaders(): Promise<Record<string, string>> {
   const headers: Record<string, string> = {
     'Content-Type': 'application/json',
   }

   try {
-    const token = await fetchDashboardToken()
-    if (token) headers.Authorization = `Bearer ${token}`
+    Object.assign(headers, await dashboardAuthHeaders())
   } catch {
-    // Token fetch is best-effort. The plugin route works without it
-    // on standard loopback installs.
+    // Dashboard auth is best-effort. The request will surface the
+    // dashboard's actual HTTP error if authentication is unavailable.
   }

   return headers
 }
```

Tested locally: before the change, Kanban returned `401 no_cookie`; after the change, it works correctly with Basic Auth.

Seems like this could be included in #758 as a small follow-up fix.

